Issue #32, opened 2015-11-05
Tooltip formatter


thorbenziemek:
It would be nice to be able to format the tooltip's contents.

coniel:
Also, it would be nice to be able to disable the tooltip.

benjycui:
@coniel do you mean that `rc-slider` should provide a prop to hide tooltip?

coniel:
@benjycui Yep, basically just expose the `noTip` prop on the Handle up to the Slider component.

just-boris:
@coniel @benjycui I have a better solution. What about prop like `handleComponent`?  By default, the bundled Handle component will be used, but it would be possible to replace to different implementation. 

It is makes sense for me, because I need to create slider with the following view:

![image](https://cloud.githubusercontent.com/assets/812240/11165985/b4d4c7b4-8b32-11e5-8324-a30ad58bb2c8.png)

Now I can do it only by patching this lib, but then I can configure handle view via configuration prop and it  makes possible to create any design of slider.

benjycui:
@just-boris It is reasonable. What is your opinio? @yiminghe @simaQ 

@coniel Next time, you should open a new issue to discuss a new feature. It is weirdo to discuss two different features in one issue. Though they are relative.

yiminghe:
ok with tipFormatter

PR #33, opened 2015-11-05
react-component/slider#32 implemented tipFormatter prop

thorbenziemek:
I've implemented the `tipFormatter` prop here. `tipFormatter` can be passed as a prop to `slider` and will be propagated to both handles. If present, the value to be formatted will be transformed using the formatter.

### DIFF
--- a/examples/simple.js
+++ b/examples/simple.js
@@ -11,6 +11,10 @@ function onChange(v) {
   console.log(v);
 }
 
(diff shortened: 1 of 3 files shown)


ande765a:
+1

benjycui:
Could you update README.md? Thanks.

simaQ:
:+1:

PR #35, opened 2015-11-13
provide slider value onAfterChange trigger

just-boris:


src/Slider.jsx
```diff
@@ -117,19 +117,7 @@ class Slider extends React.Component {
...
-      const data = {
-        upperBound: state.upperBound,
-        lowerBound: state.lowerBound,
-      };
-      data[state.handle] = value;
-      this.triggerEvents('onChange', [data.lowerBound, data.upperBound]);
-    } else {
-      this.triggerEvents('onChange', value);
-    }
+    this.triggerEvents('onChange', this.getValue());
```
benjycui: Here is a problem, see [`setState`](https://github.com/facebook/react/blob/0b29035484f428cb56e7e1c04a88f66ac020d1d4/src/isomorphic/modern/class/ReactComponent.js#L42)

benjycui:
Thanks, I will finish the rest.

PR #37, opened 2015-11-15
make range and marks props to work together

### LINKED ISSUE #36: Range mode for marks
It would be nice to have slider with specific marks in range mode.
Is there a some issues? Maybe I can help to resolve them?

just-boris:
Now you can use marks for range sliders. Added a demo and everything works fine.

closes #36

### DIFF
--- a/examples/simple.js
+++ b/examples/simple.js
@@ -9,6 +9,9 @@ var style = {width:400,margin:50};
 
 var log = console.log.bind(console);
 
+
+var marks = ["状态1","状态2","状态3","状态4"];
+
 function percentFormatter(v) {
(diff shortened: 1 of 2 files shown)


benjycui:
I had discussed with our designer, he said that there is no such use case now.

Could you show some use cases in real world?

just-boris:
It is needed for range sliders with non-linear scale, like this time range picker:
screen shot 2015-11-15 at 14 47 57

It is what I need in my project, so I am interested in this feature.

Also it is good for consistency: any option works with any other option, no exceptions.

benjycui:
:100:

PR #49, opened 2015-11-22
reduce gaps between marks

just-boris:
Now even short enough labels can't fit into one string.

I have this in my app:
screen shot 2015-11-23 at 00 19 46

With my fix, it would be better. Borders are especially have been added to show mark constraints:

screen shot 2015-11-23 at 00 34 56

### DIFF
--- a/src/Marks.jsx
+++ b/src/Marks.jsx
@@ -5,7 +5,7 @@ const Marks = ({className, marks, included, upperBound, lowerBound, max, min}) =
   const marksKeys = Object.keys(marks);
   const marksCount = marksKeys.length;
   const unit = 100 / (marksCount - 1);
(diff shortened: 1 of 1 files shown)


benjycui:
:+1:

benjycui:
IMO, this is just a temporal patch. It is still a problem in this use case:

image

What we should do is optimize the implementation of `Slider.Marks`.

just-boris:
Well, this case is more difficult. I see two options here:
1. Do not rely on marks count, use `step` value to find the appropriate mark width. For your case the minimal step is `10`, so mark width would be `10/(100 - (-10)) * 100 = 9%`
2. Make some more flexible layout, where marks may not be centered, just fill free space on any side near the dot

Issue #54, opened 2015-12-01
changing min/max modifies value


jamesjjk:
Great component! But when dynamically changing the min or max prop the value is changed? Surely it should limit to the max or min rather then changing value. Thanks!

benjycui:
Could you provide your use case? I am curious. Why do you change `min` or `max` dynamically.

jamesjjk:
Hi @benjycui. We use the component in a dynamic form, where the validation or values from other inputs affect limit the min/max value of in this case the (or ideally) the slider. For example here is a business requirement, if a user is selecting a value within a finite range which is defined by the users initial (variable input) such as maximum price or stock or units. Then the min/max will be a dependency. I hope this makes sense. Thanks for asking.

yiminghe:
what happens if current value is less than new min or greater than new max ....

jamesjjk:
Unfortunately it effects the value prop of the slider. It should only effect the value when the thing min or max is less then or more then the value respectively. A rule utilised by number pickers is:
Max < Value : Value = Max || Min > Value : Value = Min, rather then selecting a medium value between min and max.

benjycui:
@jamesjjk Forgive my poor English. I am trying to follow you.

Do you mean that:
1. Set `min=0, max=100` and the value is `10`
2. The `min` is set as `50` dynamically
3. The value should be `50`, but it is `10` now.

Right?

jamesjjk:
@benjycui Your english is ok! 
Correct and the above! AND it also sets a value when:
1. Set min=0, max=100 and the value is 20
2. The min is set as 10 dynamically
3. The value should be 20, but it is random.. now.

benjycui:
Fixed in 3.1.4

PR #55, opened 2015-12-05
fix: value should be aligned if `min` `max` changed

benjycui:
Close: https://github.com/react-component/slider/issues/54

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import {Dom as DomUtils} from 'rc-util';
 import classNames from 'classnames';
+import objectAssign from 'object-assign';
 import Track from './Track';
 import Handle from './Handle';
 import Dots from './Dots';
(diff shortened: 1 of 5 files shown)


src/Slider.jsx
```diff
@@ -214,10 +231,14 @@ class Slider extends React.Component {
     return precision;
   }
 
-  trimAlignValue(v) {
+  isAlignedByMinOrMax(value, props) {
```
just-boris: I think that `isValueOutOfBounds` is better name for function.

just-boris:
Except the little naming issue, looks good.
@benjycui :+1:

benjycui:
Not a native speaker.  囧

just-boris:
@benjycui that's fine, English is not my native language too :smile:

benjycui:
If you had reviewed this PR, and thought it is OK. Please merge it :-)

PR #58, opened 2015-12-10
chore: improve user experience #56

benjycui:
Close: #56

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -98,19 +98,16 @@ class Slider extends React.Component {
     }
   }
 
-  onChange(handle, value) {
+  onChange(state) {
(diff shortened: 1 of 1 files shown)


just-boris:
As I understand, you are going to make right handle follow to the left if it is dragged away. Now left handle just stops at the value of right handle.

Just a note, that I haven't seen this behavior before:
http://ionden.com/a/plugins/ion.rangeSlider/en.html
http://refreshless.com/nouislider/
http://jqueryui.com/slider/#range

benjycui:
Yeah, you are right. And I will add `allowCross` property to make this behaviour configurable. Users can choose what they want :-)

Issue #57, opened 2015-12-09
Basic example doesn't even work


tUrG0n:
I'm on the latest version: `3.1.4` with `react@0.14.2`

I can't get a simple basic slider to work without getting the following error:

`Cannot read property 'toFixed' of undefined`

Any clue what's this about? I need to apparently set the `marks` props... but that won't make it a basic slider right?

vgoklani:
I have the same issue, I did a fresh pull via npm and it's now broken with the same error as above :(

just-boris:
Hello! We have a demo: http://react-component.github.io/slider/ Checked this right now, it works fine. 

Is it work for you? And what is your setup of the slider? Here can be mismatches with our examples.

pkese:
Same problem here. My code is

    <Slider min={0} max={2880} defaultValue={30} value={30} />

The exception is thrown in function `trimAlignValue` with `closestPoint` = undefined:

    return step !== null ? parseFloat(closestPoint.toFixed(this.getPrecision())) : closestPoint;

just-boris:
* you cannot use value without `onChange` handler. The slider behaves like a React `input` tag. The example renders read only slider. I left only `defaultValue` and then it worked fine.
* created a jsfiddle with demo: https://jsfiddle.net/bpgtxmcp/ Does it work in your browser?

also, maybe @benjycui can help

benjycui:
@pkese Sorry, I can not re-produce that exception with your code. Could you fork @just-boris ' jsfiddle and try to re-produce it? THX.

tUrG0n:
I got my problem solved because i realised i had an problem specific to my project.
Basically, the parent container containing the slider was being redrawing twice (once before i fetch data from my api, and the when the data is fetched). At the second instance, for some reason no value was passed to the slider which was causing closestPoints to be null.
I solved it by simply refactoring my code, and having that slider redrawn once.

However, i still face the problem when i use livereload. See it loads allright the first time but then with livereload, it breaks, forcing me to do a refresh.

I think you jsfiddle is not realistic because u attach the slider to a reactdom. 
Maybe, if the demo attached the slider to a parent module, and then have some external thing going on, could help.

@benjycui Bottom line, try and put ur slider under a component that get redrawn after a fetch. I think that could reproduce the problem.

tUrG0n:
Here's another example that would break the code:

```
export default class extends Component {
  shouldComponentUpdate = shouldPureComponentUpdate;
  constructor(props) {
    super(props)
[...]

PR #60, opened 2015-12-15
fix: crash when rerender #57

benjycui:
Close: #57


just-boris:
`trimAlignValue` expects that first argument `value` is always defined. So, it is valid change.

LGTM.

PR #61, opened 2015-12-16
If tipFormatter is null, don't show the tip

frank-weindel:
We're using this component in an application where we display the actual value of the slider in a separate text box. We didn't want to show the tip in this case but there didn't seem to be any way to simply hide it. I thought setting the tipFormatter prop to null would be a sensible way to force the tip to be hidden.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -339,7 +339,7 @@ class Slider extends React.Component {
     const lowerOffset = this.calcOffset(lowerBound);
 
     const handleClassName = prefixCls + '-handle';
-    const isNoTip = tipFormatter === null || (step === null && !tipFormatter);
+    const isNoTip = (step === null) || (tipFormatter === null);
 
     const upper = (<Handle className={handleClassName}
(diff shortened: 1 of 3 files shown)


just-boris:
That feature also was requested at #32, so it seems to me a good change

src/Slider.jsx
```diff
@@ -339,7 +339,7 @@ class Slider extends React.Component {
     const lowerOffset = this.calcOffset(lowerBound);
 
     const handleClassName = prefixCls + '-handle';
-    const isNoTip = (step === null) && !tipFormatter;
+    const isNoTip = tipFormatter === null || (step === null && !tipFormatter);
```
benjycui: How about 
```js
const isNoTip = (step === null) || (tipFormatter === null);
```

benjycui:
:+1: 

Please add an example for this feature.

src/Slider.jsx
frank-weindel: Was being as careful as possible with introducing a breaking change. But I'm fine with this if you are.

frank-weindel:
Done.

Issue #63, opened 2015-12-18
feature: customizable handle


benjycui:
Lots of issues are about making tooltip more configurable.

1. #62 wants to show tooltip anytime.
1. #32 wants to format tooltip's content, and disabled tooltip.

rc-slider has become a fattie. And it seems that all of rc-tooltip's APIs are going to be exposed.

How about @just-boris ' suggestion? https://github.com/react-component/slider/issues/32#issuecomment-156748583

If rc-slider supported customizable handle, you could wrap handle with any kind of tooltip. But the drawback is that it is not as convenient as `tipFormatter` or `tipVisibility`.

just-boris:
Well, I will create a conception and example how it would look like at the next week

nilssolanki:
I'm curious, has this already been implemented?

benjycui:
Nope, I am not sure if this feature is necessary. Maybe we can try another way to solve those problems.

nilssolanki:
Do you have any workarounds at the moment?

benjycui:
I am busy with [ant-design](https://github.com/ant-design/ant-design) and [bisheng](https://github.com/benjycui/bisheng) recently. Maybe June?

You could describe your requirement(and why do you need this feature) in this issue. I will try to provide a solution.

nilssolanki:
As a matter of fact, the PR (https://github.com/react-component/slider/pull/102) looks like it addresses exactly what I need. Thank you for taking the time to look at this.

PR #67, opened 2015-12-19
Add label to Slider

Rodrigora:
Hi all!

Thanks for building this awesome Slider!

I had to add a label for the current selected value. After digging into the code, I came up with this
solution.

Please share your thoughts on this PR and LMK if you deem appropriate add this feature to Slider.

Thanks

### DIFF
--- a/assets/index.less
+++ b/assets/index.less
@@ -62,6 +62,14 @@
     }
   }
 
+  &-label {
+    position: absolute;
+    margin-top: 10px;
+    z-index: 3;
+    text-align: center;
+    font-size: 12px;
+  }
+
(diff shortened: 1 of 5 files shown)


just-boris:
Well, this is one more request for custom handles.
We are going to solve that issue in https://github.com/react-component/slider/issues/63, then you will be able to create any custom handle for your project.

Does it help you?

just-boris:
Also, it is nice to have a screenshot of new look
image

Rodrigora:
Hey @just-boris.

Yes, I think the custom handler feature would be a better choice to handle this.

PR #68, opened 2015-12-19
Allow to pass custom handle components

### LINKED ISSUE #63: feature: customizable handle

just-boris:
First attempt to resolve #63.

I like that because it doesn't cause a lot of changes in the source code. It just works. Then we need to document it and describe the properties which are passed to handle component.

I tried to make few examples how it can be used. This is how it looks like.
image

### DIFF
--- a/examples/custom-handles.html
+++ b/examples/custom-handles.html
@@ -1,0 +1,1 @@
+placeholder
--- a/examples/custom-handles.js
+++ b/examples/custom-handles.js
@@ -1,0 +1,58 @@
+require('rc-slider/assets/index.less');
+
+const React = require('react');
+const ReactDOM = require('react-dom');
+const Slider = require('rc-slider');
+
+const style = {width: 400, margin: 50};
+
+const handleStyle = {
+  position: 'absolute',
+  transform: 'translate(-50%, -50%)',
+  cursor: 'pointer',
+  padding: '2px',
+  border: '2px solid #abe2fb',
+  borderRadius: '3px',
+  background: '#fff',
+  fontSize: '14px',
+  textAlign: 'center',
+  zIndex: 3
+};
+
+class CustomHandle extends React.Component {
+  render() {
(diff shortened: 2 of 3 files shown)


yiminghe:
just pass a handle element is more customizable .

```js
<Slide handle={<MyHandle xx="yy"/>}>
```

just-boris:
@yiminghe it will not work. You can't get handle position in that way.

If you want to pass extra props to custom handle, you can do this

```
<Slide handle={({offset, value}) => <MyHandle xx="yy" offset={offset}>}>
```

And technically it would be a short form of this:

```
function HandleWrapper({offset, value}) {
   return <MyHandle xx="yy" offset={offset}>;
}
<Slide handle={HandleWrapper}>
```

yiminghe:
inside rc-slider:
```
React.cloneElement(handle, {
 ...
});
```

see usage of rc-calendar: http://react-component.github.io/calendar/examples/antd-calendar.html

just-boris:
As it seems to me, it is a bit unclear for developer how it works.

But since you already have success story in the calendar, so it should be done here in the same way. I will update my PR by the end of week.

benjycui:
Close for no update. 

Just reopen or create a new PR when you have time to complete it. @just-boris

just-boris:
Sorry, I messed up and forgot about this. Of course, I will make all necessary fixes and create new when I got the time.

PR #70, opened 2016-01-26
ReverseSlide function added, small Dots bug fixed

ataipale:
ReverseSlide: Allows the slider to be tethered on the right and dragged
from the left.
Dots bug: When `step=null` while `dots=true` the Dots code breaks your
browser by trying to add null. Also, dots are added even when
`dots=false` if you pass marks as a param to the slider. I have fixed
both of these issues.


benjycui:
Sorry, confict ..

And the following is expected behaviour..
>  Also, dots are added even when dots=false if you pass marks as a param to the slider. I have fixed
both of these issues.

PR #71, opened 2016-02-02
put marks in correct order so one can style it

kalinichenko:
I need to change styles of the first and last marks. I couldn't find any other way but to put them in correct order.

### DIFF
--- a/src/Marks.jsx
+++ b/src/Marks.jsx
@@ -8,7 +8,7 @@ const Marks = ({className, marks, included, upperBound, lowerBound, max, min}) =
   const markWidth = unit * 0.9;
 
   const range = max - min;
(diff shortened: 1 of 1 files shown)


just-boris:
Hello. Why can't you sort marks before passing it down to the component?

Also I have a use case with string values in marks: 

image

Your pull request probably will break this use-case.

kalinichenko:
@just-boris I am not sure I got you right. Slider expects to get an object for marks. Marks component use keys from marks object to create elements. How can I specify an order for them?

just-boris:
Well, I got it. Now you have the following collection:

```js
var marks = {
   2: 'c',
   0: 'a',
   1: 'b'
}
```
And you want to preserve the order of keys. That makes sense.

LGTM.

benjycui:
THX!

PR #77, opened 2016-02-29
multi touch support

lukephills:
Hi,

Currently when getting touch position the slider gets the position of first touch which breaks multi touch apps.

I've updated to get the position of the specific touch event so that it doesn't override other touches functionality.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -13,7 +13,7 @@ function noop() {
 }
 
 function getTouchPosition(e) {
-  const touches = e.changedTouches;
+  const touches = e.touches;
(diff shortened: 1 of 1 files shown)


yiminghe:
_touchIdentifiers should be instance field?

src/Slider.jsx
```diff
@@ -250,6 +250,8 @@ class Slider extends React.Component {
     return precision;
   }
 
+  touchIdentifiers = {};
```
benjycui: Syntax Error?

benjycui:
Could you update unit test?

benjycui:
Close for no response for a lone time.

Issue #76, opened 2016-02-29
Mark labels overflow slider width


vdh:
Due to the way the labels are positioned with `position: absolute` and percentage widths, the labels are very often overflowing very badly off the left and right edges of their container, especially if very long label text is used.

just-boris:
Confirm that issue, but I don't see here an easy fix.
As I understand, we need to compute `markWidth` [here](https://github.com/react-component/slider/blob/master/src/Marks.jsx#L8) for each mark regarding its content. Also, because of React Virtual DOM we don't have access to real DOM-nodes to get actual sizes of marks.

So, we need to refactor here and arrange marks on `componentDidMount`, where we can get their actual DOM-nodes and arrange them properly.

@vdh if you will try to something in this way, it will be much appreciated.

just-boris:
Also, we have a got a pull-request #78 that possibly can fix your issue

benjycui:
3.4.0

Issue #50, opened 2015-11-23
feat: the width of mark should be configurable


benjycui:
See: #49

benjycui:
3.4.0

PR #78, opened 2016-02-29
Added ability to override mark styling/positioning.

aweber1:
I had a need to modify the positioning of the min/max marks, which is applied as an inline style and not override-able via stylesheet classes. So I implemented a simple callback that will allow you to specify/calculate/override the inline styles for individual marks. Should help with #76 and #50. 

Example:
```js
function markStyler(index, marksCount, point, defaultStyle){
    if (index === 0){
      return {width: defaultStyle.width, left: 0};
    }
    if (index === marksCount - 1){
      return {width: defaultStyle.width, right: 0, left: "initial"};
    }
    return defaultStyle;
  }

```html
<Slider min={-10} marks={marks} step={null} onChange={log} defaultValue={20} markStyler={markStyler} />
```

Thanks for the great component!

### DIFF
--- a/src/Marks.jsx
+++ b/src/Marks.jsx
@@ -1,14 +1,14 @@
 import React from 'react';
 import classNames from 'classnames';
 
-const Marks = ({className, marks, included, upperBound, lowerBound, max, min}) => {
+const Marks = ({className, marks, included, upperBound, lowerBound, max, min, markStyler}) => {
   const marksKeys = Object.keys(marks);
   const marksCount = marksKeys.length;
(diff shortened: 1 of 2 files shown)


benjycui:
I don't like inline style, but it seems that there is no better solution @yiminghe

yiminghe:
```js
marks: {
 0: {
  style: yyy
  children: xxx
},
1: zzz
}
```

benjycui:
@aweber1 Config object is simpler than function, see: https://github.com/react-component/slider/pull/82

PR #82, opened 2016-03-17
feat: support set style of specific mark point

benjycui:
Close: https://github.com/react-component/slider/issues/50 https://github.com/react-component/slider/issues/76


just-boris:
very great!
thanks @benjycui

benjycui:
3.4.0

PR #83, opened 2016-03-17
fix marks only one bug

maolion:


benjycui:
Could you fix those ESLint errors? And add a unit test for one mark.

src/Marks.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 const Marks = ({className, marks, included, upperBound, lowerBound, max, min}) => {
   const marksKeys = Object.keys(marks);
   const marksCount = marksKeys.length;
-  const unit = 100 / (marksCount - 1);
+  const unit = 100 / ((marksCount - 1)||1);
```
just-boris: Maybe `Math.min(marksCount - 1, 1)`

benjycui: `Math.min(1 - 1, 1)` => 0

just-boris: I meant `Math.max(marksCount - 1, 1)`, sorry

benjycui:
Close for no response for a long time.

PR #91, opened 2016-04-14
avoid infinite loop when passing in dots with no step

Kooshaba:
I mistakenly passed in dots with no steps and it caused an infinite loop while rendering.

I know you're not supposed to use this config, but I feel this should be avoided altogether.


benjycui:
A warning will be better.

PR #95, opened 2016-04-20
using actual step value instead of outdated prop.

franckamayou:


benjycui:
This PR cannot solve any problem, just open an issue to show how to re-produce the bug.

Issue #97, opened 2016-04-26
Cannot move handles


nwehrle:
### Example
`<RcSlider range={true} min={0} max={40} defaultValue={[10, 10]}/>`

### Expected
Both handles overlap eachother. I can touch and move a handle to the left or right.

### Actual
I cannot move any of the handles.

### Videos
Component:
![component](https://cloud.githubusercontent.com/assets/832234/14826880/282ba9f2-0be0-11e6-9d63-e4f4697481cb.gif)

DevTools: 
![devtools](https://cloud.githubusercontent.com/assets/832234/14827042/c421a898-0be0-11e6-87d8-852b772d9f48.gif)

### Versions
rc-slider@3.5.1

benjycui:
3.6.2

PR #99, opened 2016-04-27
Fix #96 and #97

### LINKED ISSUE #96: Slider crashes when nextProps.null !==  null and this.props.step === null.
This issue happens because the function getPrecision is reading this.props which is not up to date yet.
To reproduce the error, Render slider with step={null} then re-render with step={1}

benjycui:


src/Slider.jsx
```diff
@@ -48,8 +48,7 @@ class Slider extends React.Component {
     if (props.range && upperBound === lowerBound) {
       if (lowerBound === max) {
         recent = 'lowerBound';
-      }
-      if (upperBound === min) {
+      } else {
         recent = 'upperBound';
       }
```
just-boris: maybe

```js
recent = lowerBound === max ? 'lowerBound' : 'upperBound';
```

would be better?

just-boris:
Very good.

benjycui:
Updated.

just-boris:
👍 good job, @benjycui

PR #101, opened 2016-05-05
Add license file

nathandurand:
I tried to add rc-slider to WebJars ( http://www.webjars.org/npm ) to be used from Java / Scala-js but it failed with:

`Failed! All attempts to determine a license have been exhausted. The bower.json did not contain a spec-compliant license definition and the license could not be determined by trolling through the GitHub repo. This problem will likely need to be resolved by working with the library maintainers directly. If you feel you have reached this failure in error, please file an issue: https://github.com/webjars/webjars/issues`

I hope adding a LICENSE file will solve this issue.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 The MIT License (MIT)
-Copyright (c) 2015-2016 Benjy Cui
+Copyright (c) 2015-present Alipay.com, https://www.alipay.com/
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
+Copyright (c) 2015-2016 Benjy Cui
```
benjycui: Should be Alipay, see: https://github.com/ant-design/ant-design/blob/develop-1.0.0/LICENSE

nathandurand: No worries, fixed.

PR #102, opened 2016-05-17
Allow to pass custom handle components

elboman:
This is my attempt to solve #63  starting from #68 

It's basically the same, I changed the API in order to pass the custom handle directly as prop, using internally `cloneElement` instead of the function, as suggested by @yiminghe 

As shown in the example:
```javascript
<Slider Handle={<MyHandle />} />
```
which I agree is much more customizable

Let me know, and in case, I guess we just need to update the docs


benjycui:
`Handle` => `handle`

It's just an element, not a function.


And could you add this API to README.md? Thanks.

benjycui:
3.7.0

PR #116, opened 2016-07-07
Update Slider.jsx

alanquigley:
Fixed a bug that caused the sliders to break when passed values such as min=70 max=310 step=20


benjycui:
Fixed in https://github.com/react-component/slider/commit/e2504d99fe845d686ac468279b5df5008cc8ed04 , and thanks! 

Next time, please fix ESLint errors before you create a pull request.  😀

benjycui:
3.7.4

PR #117, opened 2016-07-12
Allow custom track and step renderers.

alampros:
Add `trackComponent` prop.
Add `stepsComponent` prop.
Add tests and updated docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,39 @@ ReactDOM.render(<Rcslider />, container);
           <td>When the `step` value is greater than 1, you can set the `dots` to  `true` if you want to render the slider with dots.</td>
         </tr>
         <tr>
+          <td>trackComponent</td>
+          <td>Component</td>
+          <td></td>
+          <td>Provide a custom track component that receives the following props:
+            <ul>
+              <li><code>className</code>: <em>String</em></li>
+              <li><code>included</code>: <em>Boolean</em></li>
+              <li><code>vertical</code>: <em>Boolean</em></li>
+              <li><code>offset</code>: <em>Number</em></li>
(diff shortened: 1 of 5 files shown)


just-boris:
Hello! Thank you for you contribution!

But track and step are just a single html-element. Why do you need to customize them? I think, that the most customizing issues can be solved with just CSS.

alampros:
@just-boris No problem, thanks for the quick response!

I'd like to provide a custom track component that draws a histogram of all available values using a `<canvas>` element. This would give the user some idea of what kind of change their actions might produce. I'll post a screenshot of my mockup when it's done.

Feel free to close this PR if this is too much of an edge case.

just-boris:
Well, I saw similar example in [angular-slider](http://angular-slider.github.io/angularjs-slider/) projejct (see "Slider with dynamic selection bar color" demo). Maybe it makes sense.

The final decision is on @benjycui

alampros:
Hold off on merging this – I think I botched the implementation. I'll rethink it and reopen it if needed. Sorry!

just-boris:
Well, just use [custom handle implementation](https://github.com/react-component/slider/blob/master/src/Slider.jsx#L342) as example.

alampros:
I think I got it. Sorry, I should be allowed to push commits without sufficient coffee.

benjycui:
> Feel free to close this PR if this is too much of an edge case.

Yes, it is not common use case.

> I'll post a screenshot of my mockup when it's done.

This PR will be close, but I still hope that you could post your screenshot. If someone need similar feature, maybe he/she can use your fork version.

Thanks :-)

PR #118, opened 2016-07-14
Identify upper/lower range handles

miskreant:
Extend the handle classNames, and provide props so that
upper/lower handles are easily distinguishable in CSS & React.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -337,15 +337,19 @@ class Slider extends React.Component {
     const lowerOffset = this.calcOffset(lowerBound);
 
     const handleClassName = prefixCls + '-handle';
+
+    const upperClassName = handleClassName + ' ' + handleClassName + '-upper';
+    const lowerClassName = handleClassName + ' ' + handleClassName + '-lower';
+
     const isNoTip = (step === null) || (tipFormatter === null);
 
-    const upper = cloneElement(customHandle, { className: handleClassName,
+    const upper = cloneElement(customHandle, { className: upperClassName,
         noTip: isNoTip, tipTransitionName: tipTransitionName, tipFormatter: tipFormatter,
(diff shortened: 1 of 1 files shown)


src/Slider.jsx
```diff
@@ -337,16 +337,20 @@ class Slider extends React.Component {
...
+
+    const upperClassName = handleClassName + ' ' + handleClassName + '-upper';
+    const lowerClassName = handleClassName + ' ' + handleClassName + '-lower';
+
     const isNoTip = (step === null) || (tipFormatter === null);
 
-    const upper = cloneElement(customHandle, { className: handleClassName,
-        noTip: isNoTip, tipTransitionName: tipTransitionName, tipFormatter: tipFormatter,
+    const upper = cloneElement(customHandle, { className: upperClassName,
+        noTip: isNoTip, tipTransitionName: tipTransitionName, tipFormatter: tipFormatter, upperHandle: true, lowerHandle: false,
```
benjycui: `className` is enough, just remove `upperHandle` and `lowerHandle` to simplify code.

src/Slider.jsx
```diff
@@ -337,16 +337,20 @@ class Slider extends React.Component {
     const lowerOffset = this.calcOffset(lowerBound);
 
     const handleClassName = prefixCls + '-handle';
+
+    const upperClassName = handleClassName + ' ' + handleClassName + '-upper';
```
benjycui: Use `classNames` instead of string concatenation.

benjycui:
Thanks!

miskreant:
Updated PR with feedback, thanks for the suggestions!

src/Slider.jsx
miskreant: Removed them. You might still consider it though, as it makes identifying an upper/lower handle a little easier from within a custom handle component. Otherwise you'll have to search the className string for 'upper' or 'lower'.

benjycui:
3.8.0

PR #122, opened 2016-07-21
Ranges: Make the value of the opposing handle available.

miskreant:
This allows more flexibility when implementing a customHandle
by allowing you to use the position of the lowerbound while
calculating the offset of the upperbound, and vice versa.

Some example use cases: 
* Upper and lower bounds should always be X distance apart
* Upper/lower handle should stick to the range end when the other is less/greater than X
* One handle should change style based on the proximity of the other

### DIFF
--- a/src/Handle.jsx
+++ b/src/Handle.jsx
@@ -57,6 +57,7 @@ Handle.propTypes = {
   tipTransitionName: React.PropTypes.string,
   tipFormatter: React.PropTypes.func,
   value: React.PropTypes.number,
+  otherValue: React.PropTypes.number,
   dragging: React.PropTypes.bool,
   noTip: React.PropTypes.bool,
 };
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -352,13 +352,15 @@ class Slider extends React.Component {
 
     const upper = cloneElement(customHandle, { className: upperClassName,
         noTip: isNoTip, tipTransitionName: tipTransitionName, tipFormatter: tipFormatter,
-        vertical: vertical, offset: upperOffset, value: upperBound, dragging: handle === 'upperBound' });
+        vertical: vertical, offset: upperOffset, value: upperBound, otherValue: (range ? lowerBound : null),
+        dragging: handle === 'upperBound' });
 
     let lower = null;
     if (range) {
       lower = cloneElement(customHandle, { className: lowerClassName,
         noTip: isNoTip, tipTransitionName: tipTransitionName, tipFormatter: tipFormatter,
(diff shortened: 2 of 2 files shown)


benjycui:
You can get other value like this:

```jsx
<Slider range handle={<Handle value={[0, 100]} />} />
```

And, you can identify handle with it's `className`, see: https://github.com/react-component/slider/pull/118

miskreant:
Thanks for the suggestion, however that will only work if the entire Slider is re-rendered whenever the handle moves. If the parent component is a stateless component, it will not send the rangeValue props down to the Handle.

Consider the example below: 

```JS
/* eslint react/no-multi-comp: 0 */
require('rc-slider/assets/index.less');

const React = require('react');
const ReactDOM = require('react-dom');
const Slider = require('rc-slider');

const style = {width: 400, margin: 50};

const handleStyle = {
  position: 'absolute',
  transform: 'translate(-50%, -50%)',
  cursor: 'pointer',
  padding: '2px',
  border: '2px solid #abe2fb',
  borderRadius: '3px',
  background: '#fff',
  fontSize: '14px',
  textAlign: 'center',
  zIndex: 3,
};

const CustomHandle = props => {
  // The handle does not get the new rangeValue unless the entire slider is re-rendered
  window.console.log('Handle Value', props.rangeValue);
  const style = Object.assign({left: props.offset + '%'}, handleStyle);
  return (
    <div style={style}>val: {props.value}</div>
  );
};
CustomHandle.propTypes = {
  value: React.PropTypes.any,
  offset: React.PropTypes.number,
};

const CustomizedRange = React.createClass({
  getProps: function() {
    const low = Math.random() * 100;
    const high = Math.random() * 100 + low;
    return {
      range: true,
      allowCross: false,
      value: [low, high],
    };
  },
  onSliderChange: function(value) {
    // The Slider value has changed, but the CustomHandle will be unaware
    window.console.log('Range Value', value);
    // Unless, you re-render the entire slider
    // this.forceUpdate(); // This should not be necessary
  },
  render: function() {
    const props = this.getProps();
    window.console.log('Render Slider');
    return (
      <div>
        <Slider {...props} onChange={this.onSliderChange} handle={<CustomHandle rangeValue={props.value} />} />
      </div>
    );
  },
});

ReactDOM.render(
  <div>
    <div style={style}>
      <p>Customized Range</p>
      <CustomizedRange />
    </div>
  </div>
  , document.getElementById('__react-content'));

benjycui:
I don't like `this.forceUpdate`, either.

Use `Slider` as a controlled component, everything will be better :-)

`CustomizedRange` see: http://react-component.github.io/slider/examples/range.html

PR #127, opened 2016-07-27
Add 'style' node to package.json

heldr:
Many modules in npm are starting to expose their css entry files in their `package.json` files. This allows tools like `postcss-import`, `parcelify`, `npm-css`, `rework-npm` and `npm-less` to import *rc-slider* straight from node_modules directory.

Example:
```css
@import 'rc-slider';
```

The same way as `main` node, but for css 😄 

Reference:
https://github.com/twbs/bootstrap/pull/12441

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   ],
   "licenses": "MIT",
   "main": "./lib/index.js",
+  "style": "./assets/index.css",
   "config": {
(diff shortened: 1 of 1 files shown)


benjycui:
Cool, thanks!

benjycui:
3.8.2

PR #130, opened 2016-07-29
Add point value as classname for marks

TomiS:
Hi there and thanks for the nice library.

I propose point value to be added as a name spaced class to each Mark in Marks.jsx. This enables easier targeting of css styles to each mark individually. Would you agree?


benjycui:
These className doesn't make sense while point's value is float(e.g. `1.1`).

You can use marks with style directly, see: http://react-component.github.io/slider/examples/marks.html

Anyway, thank you for your contribution. 😄

PR #133, opened 2016-08-06
Add multi-range support

sosz:
Added support for more than 2 handles using `<Slider range={<number>} />`.
`number` is the number of tracks (i.e., `number === nHandles + 1`).

Example of `<Slider range={2} value={[25, 50, 65]} />`:

multi-range
>CSS is different because of my application custom stylesheets...

### Pushable Handles
I also need pushable handles in my webapp...

The `pushable` property can be set to `true` to enable handle-pushing. When it is a number, the number represents the minimum distance between handles when pushing (`pushable={true}` is equivalent to `pushable={1}`).

Example of `<Slider range={3} pushable={10} defaultValue={[10, 25, 45, 70]} />`:

pushable-handles

It also works correctly with any weird combination of `marks` and `step`.

### Implementation Details
- Instead of `lowerBound` and `upperBound`, the `Slider`'s state was changed to store a `bounds` array.
- Pushable handles calculates beforehand (and caches) all possible handle positions, taking into account `marks` and `step`. It then tries to push the surrounding handles away from the moving handle until the needed threshold (the value of `pushable`) is satisfied.

### IMPORTANT
I couldn't manage to run the test suite. I've cloned the original `slider` repository. Executed:
```
npm install
npm start
```
But when I go to *http://localhost:8005/examples/*, I'm missing resources (they are being searched in the wrong place...). It's the same for tests.

I've only run manual tests on my webapp.

I did modify the test suite, and also added some tests for the new multi-range, but someone must run the test suite to ensure everything works...

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ slider ui component for react
 
 <img src="https://t.alipayobjects.com/images/T1wO8fXd4rXXXXXXXX.png" width="550"/>
 
+<img src="http://i.giphy.com/l46Cs36c9HrHMExoc.gif"/>
+
 
 ## Feature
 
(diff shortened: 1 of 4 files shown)


benjycui:
It is a complicated PR, I need time to review it, thanks.

benjycui:
👍

benjycui:
4.0.0

rcsole:
Since you bumped a major version, does this have any breaking changes
or are you expecting there to be any?

benjycui:
Nope... You can consider that I make a mistake...

Issue #135, opened 2016-08-15
Z-index bugs


derrickpelletier:
Because the child components (track, handles, etc) are all arranged with z-indexes, they break when used with other layering type components like `react-modal`, `react-select`, etc. The slider will appear _over_ the modal or select box in these cases.

I removed the z-indexes in the stylesheet and re-arranged the components so they would layer appropriately by order.

PR #136, opened 2016-08-15
Removed z-indexes #135

derrickpelletier:
For #135 

- Removed z-indexes from styles.
- Rearranged components
- updated example that used z-indexes


derrickpelletier:
Not sure what the deal is with the failed travis checks but appears unrelated.

benjycui:
4.0.1

PR #137, opened 2016-08-15
Fix issue where invalid bounds would be calculated

dsgkirkby:
In some cases, specifically when max and min aren't clean multiples of the step, the bounds can be set to invalid values (> max). The test case added in this PR fails on the current version.

This is a big problem because a controlled slider that encounters this bug will enter a near-infinite (this bug caused call stack overflow errors in our application) render loop as it continually receives, and passes down, new values that are still invalid.

The issue is that the call to Math.round [here](https://github.com/react-component/slider/blob/master/src/Slider.jsx#L277) can push the value above the max.
For example, if max is 2.1, min is 0.5, and step is 1, then the expression `(val - min) / step` is `(2.1 - 0.5) / 1 === 1.6`, which is then rounded up to 2 (and then the min is added, giving us a new value of 3, which is above the max).

 To resolve this, instead of using the given maximum as the upper limit, I've used the closest smaller value that could be selected (i.e. that is an integer multiple of step above the min). I believe this is the least disruptive means of resolving this bug.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -260,8 +260,9 @@ class Slider extends React.Component {
     if (val <= min) {
       val = min;
     }
-    if (val >= max) {
-      val = max;
+    const maxSelectableValue = (Math.floor((max - min) / step) * step) + min;
+    if (val >= maxSelectableValue) {
+      val = maxSelectableValue;
     }
     /* eslint-disable eqeqeq */
     if (!allowCross && handle != null && handle > 0 && val <= bounds[handle - 1]) {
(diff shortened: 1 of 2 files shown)


benjycui:
So, in your application, `min` and `max` are dynamic, right? 

For rc-slider expects that `(max - min) % step === 0`, and it is addressed in README.md....

> max - min should be evenly divisible by the step value.

benjycui:
You can try to rebase, so that the CI may success..

dsgkirkby:
Yes, they're dynamic.
I missed that in the readme, but I've fixed the behaviour in our app by making the step smaller so that it would evenly divide.
If you don't want to try to cover this case, then it probably doesn't make sense to merge this PR. 
I think it would probably be good to add a warning in dev if that invariant is broken though. If this is something you think would be a good idea, I'd be happy to take a stab at another PR for that.

benjycui:
> add a warning

👍 

And thanks.

PR #139, opened 2016-08-16
Add `prefixTrack` property for range sliders

sosz:
Range sliders render a track between every two handles. I've added a `prefixTrack` property (defaulted to `false`) that enables adding a track from offset 0 to the first handle (with css classes `track track-0 track-prefix`).

Also in this pull request:
- Fixed multi-range example (to use `defaultValue` instead of `value`)
- Add `pushable` example
- Add `prefixTrack` example

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,12 @@ ReactDOM.render(<Rcslider />, container);
           <td>Determines the type of slider. If range is `true`, two handles will be rendered in order to select a range. If range is a number, multiple handles will be rendered (number + 1). Using `range={true}` is equivalent to `range={1}`.</td>
         </tr>
         <tr>
+          <td>prefixTrack</td>
+          <td>boolean</td>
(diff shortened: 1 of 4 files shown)


benjycui:
I don't understand the use case of `prefixTrack`, could you show me the use story?(instead of functionality)

sosz:
I personally need it for the following:

My users need to select 3 sections on the range from 0% to 100%
```
[===========O===========O===========O------]
```
The first section always starts from 0%. So that's why I want an initial track.

I color each section (using CSS). 

A better, more generic solution could be to have unmovable handles at set locations.

benjycui:
How about http://react-component.github.io/slider/examples/custom-handles.html

If it is the first handle, just disabled it?

For this is really an edge case..

benjycui:
And could you rebase and fix CI errors if any?

PR #140, opened 2016-08-17
Colors for Tracks and Handles

soledaddiez:
I have done some changes on demand for my project. 
Just wondering if you are interested in merging it back.
Open to discusion / suggestions.

You can now define marks as
```
var marks = {
      0: { style: {}, label: 'Zero', color: '#fff' },
      1: { style: {}, label: 'One', color: '#234' },
      2: { style: {}, label: 'Two', color: 'red' },
      3: { style: {}, label: 'Three', color: 'black' },
      4: { style: {}, label: 'Four', color: '#783390' }
    };
```
If `color` is selected: 
- Handle will take `color` when positioned over the mark
- Track will be coloured with a linear gradient from `color` in mark one to `color` in mark 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,9 +94,9 @@ ReactDOM.render(<Rcslider />, container);
         </tr>
         <tr>
           <td>marks</td>
-          <td>object{number: string} or object{number: object{ style, label }}</td>
+          <td>object{number: string} or object{number: object{ style, label, color }}</td>
           <td>{}</td>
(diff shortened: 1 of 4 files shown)


benjycui:
Just override stylesheet, thanks.

soledaddiez:
@benjycui I haven't seen a way to override the track style dynamically depending on the selected marks. But I may be wrong according with what you say. Any example?.

PR #142, opened 2016-08-19
Add warning when range isn't a multiple of the step

dsgkirkby:
As per https://github.com/react-component/slider/pull/137, added a warning when range isn't a multiple of the step.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -7,7 +7,7 @@ import Steps from './Steps';
 import Marks from './Marks';
 import warning from 'warning';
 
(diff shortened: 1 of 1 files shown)


benjycui:
Sorry for late response.

Could you rebase and fix CI errors if any?

And I think you can just add this warning in `constructor`, and the condition may be this:

```jsx
if (process.env.NODE_ENV !== 'production' && (max - min) % step !== 0) {
  ...
}
```

dsgkirkby:
@benjycui done

PR #145, opened 2016-08-24
fixed `onChange` which fired on every mousmove, added test

Fuzzyma:
This PR fixes react-component/slider/issues/144

I tried to lint but linting fails because of modules not installed and wrong config values.
Also tried using `ReactTestUtils.Simulate` for tests without success.
At least I tried...


benjycui:
Thanks!

But, could you resolve CI failed?

Fuzzyma:
At the time writing this PR your test suite failed because of an internal error and not because of the actual test failing. Is that fixed?

benjycui:
Sorry, my bad..

benjycui:
Done, please rebase, thanks!

PR #148, opened 2016-09-06
Support custom class in Marks

TomiS:
This PR adds support for custom classes in Marks.jsx. It allows customization of each mark using css without using inline styles. It should not break BC according to my quick testing :)

Would you consider this worth adding?

In practice, this allows developer to write something like:

```
    let marks = {};
    for (let i = 1; i <= 20; i++) {
      let color = (i % 2 == 0) ? 'green' : 'red';
      marks[i] = {
        className: 'mark-color-'+color,
        label: i
      }
    }
```

and produce Mark elements like this

```
<span class="rc-slider-mark-text mark-color-green" style="...">4</span>
```


TomiS:
I think this conflicts because I actually implemented this month ago to a slightly older version. I can recreate this to the tip of master if you approve the idea and are interested in merging.

benjycui:
Could you rebase and fix CI errors if any?

PR #156, opened 2016-09-28
Add support for customizing a mark label in the :hover state

dstarcev:
When the mouse pointer is over a dot on the slider, add `rc-slider-mark-text-hover` class to the corresponding label.
This can't be done with CSS as the dot and text tags are not related in the DOM.
This may be useful when you want marks to react somehow to mouse movements, e.g. show/hide labels

### DIFF
--- a/src/Marks.jsx
+++ b/src/Marks.jsx
@@ -1,7 +1,8 @@
 import React from 'react';
 import classNames from 'classnames';
 
-const Marks = ({ className, vertical, marks, included, upperBound, lowerBound, max, min }) => {
+const Marks = ({ className, vertical, marks, included, upperBound,
+                 lowerBound, max, min, marksHover }) => {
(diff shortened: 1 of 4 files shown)


benjycui:
Maybe the better way is to refactor and put point and mark in the same parent node？

dstarcev:
I agree with you, it would be better from the design point of view. I had the same thoughts when I was making this pull request. However, it would be a major change, probably breaking change for those who had customized the slider's CSS.
I suggest that we include this patch into one of the next minor releases and I create another pull request with the needed refactoring for a major release that will come later. What do you think guys?

benjycui:
Just refactor and we can publish a major version.

dstarcev:
Ok, will do it.

benjycui:
Close now, and waiting for a new PR

PR #157, opened 2016-10-04
add sass version style file

ghostuz:
I have manual convert less to sass(scss) versoin.
BTW. this scss version depend on Compass
我人工将 less翻译成了 sass(scss)版本, 引用了 Compass组件


benjycui:
Just keep it simple.

We have to plan to switch from less to sass recently, for all the components in react-component and ant-design.

PR #167, opened 2016-11-04
Make the second argument to tipFormatter the index of the handle

rzane:
This is useful when working with an array of objects:

```jsx
const items = [{ name: 'A', value: 25 }, { name: 'B', value: 60 }];
const tipFormatter = (value, index) => items[index].name;

<Slider
  value={items.map(item => item.value)}
  tipFormatter={tipFormatter}
/>
```

By the way, I really wish that this package could be installed as a git source. I totally understand the desire to keep build files out of the repo, but not committing `lib` is very inconvenient. For me to use the fork I've created, I basically have to vendor the entire project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ ReactDOM.render(<Rcslider />, container);
           <td>tipFormatter</td>
           <td>function or `null`</td>
           <td></td>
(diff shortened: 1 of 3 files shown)


benjycui:
lint failed.

And.. could you update the documentation for `tipFormatter` at the same time?

rzane:
All done. Sorry for taking so long!

PR #169, opened 2016-11-14
fix slider start position

yshnb:
When magnifying screen, I cannot touch and move handles on the slider correctly.

and found it:
```es6
   getSliderStart() {
     const slider = this.refs.slider;
     const rect = slider.getBoundingClientRect();
 
     return this.props.vertical ? rect.top : rect.left;
```

and

```es6
   calcValueByPos(position) {
     const pixelOffset = position - this.getSliderStart();
```

https://developer.mozilla.org/en/docs/Web/API/Element/getBoundingClientRect
> This means that the rectangle's boundary edges (top, left, bottom, and right) change their values every time the scrolling position changes (because their values are relative to the viewport and not absolute). If you need the bounding rectangle relative to the top-left corner of the document, just add the current scrolling position to the top and left properties (these can be obtained using window.scrollX and window.scrollY) to get a bounding rectangle which is independent from the current scrolling position.

position is absolute value, so it is probably required to compensate for original values.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -249,7 +249,7 @@ class Slider extends React.Component {
     const slider = this.refs.slider;
     const rect = slider.getBoundingClientRect();
 
-    return this.props.vertical ? rect.top : rect.left;
+    return this.props.vertical ? (window.scrollY + rect.top) : (window.scrollX + rect.left);
   }
 
   getPrecision(step) {


benjycui:
Yep, it seems that you are right.

But, it curious that I can re-produce it with this demo http://react-component.github.io/slider/examples/slider.html

I had zoom the page, but it works fine. Could you provide a procedure to re-produce it?

yshnb:
I'm sorry that my explanation was not sufficient.

I found it on smartphone with zooming, and additionally, I had checked it when range option is enable. I guess, it was produced because two handles are displayed.

please see the below example on smartphone, and try zooming it.
http://react-component.github.io/slider/examples/range.html

yshnb:
e.g. https://gyazo.com/f8d6495c255f7f836df4a9a0354bc6e5

yshnb:
Sorry, my using package is a little older version, and latest version is not so much problematic.

PR #172, opened 2016-11-16
extended with min and max values

aybb:


benjycui:
CI failed.

And please address what are you going to fix.

aybb:
I was looking a way to extend slider with min and max values user can chose. While range don't changes, i need to control those values. Anyway i reached this by controlling values from my code. Thanks.

benjycui:
So, close now.

PR #180, opened 2016-11-30
Allow to align the tooltip with the slider. Also allow to set the html container for the tooltip.

mlipouski:
This allows to align the tooltip with the slider. It takes advantage of the `align` prop of the [rc-tooltip](https://github.com/react-component/tooltip) dependency.

Also I found the case where it would be useful to set the html container for the tooltip. I came across the problem when tooltip doesn't display correctly relating to the slider element. It occurred when the container of the slider is scrollable and doesn't match the tooltip container. When scrolling, slider element moves whereas tooltip stays in the same place. The issue arises because the `rc-tooltip` appends tooltip to the `body` by default. The ability to allow to set the container for the tooltip will solve the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,12 @@ ReactDOM.render(<Rcslider />, container);
           <td>Align the tooltip with the slider.</td>
         </tr>
         <tr>
+          <td>tipContainer</td>
+          <td>function</td>
+          <td></td>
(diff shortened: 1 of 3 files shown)


benjycui:
Just use customize handle now.

rc-slider need to be refactor and provide `Slider[showTooltip]`, trace: https://github.com/react-component/slider/issues/198

PR #181, opened 2016-11-30
added tipAlwaysVisible prop

mlipouski:
This allows the tooltip to be always visible.


aderushev:
+1

benjycui:
Just use customize handle now.

rc-slider need to be refactor and provide Slider[showTooltip], trace: #198

Issue #189, opened 2016-12-14
Bump rc-utils version


FallingSnow:
Can you please bump the rc-utils version to 4.0.2, I think it will fix an error I'm getting.

benjycui:
A RP is welcomed, thanks.

FallingSnow:
Can I just edit the file and make a request or do I need to build the files too.

benjycui:
You should test locally before you create a PR

FallingSnow:
#191

PR #191, opened 2016-12-17
Update rc-utils to 4.0.0

FallingSnow:
Close: #189


STRML:
Please merge this; `rc-trigger` calls `this.renderComponent` which is provided by rc-util's `getContainerRenderMixin`, but `getComponent` as called in that function does not resolve because it is defined directly on `<Trigger />`, and not passed to the mixin.

As a result, a fresh install of `rc-slider` does not work.

PR #197, opened 2016-12-27
General fixes and improvements

simeg:


just-boris:
Hello! Even though we are appreciating your work, could you give us a few words about the motivation behind this change?

As I can see, it is mostly about code style, not any specific functionality.

simeg:
Hi,

Sure, I should've done that from the start. Some of the commits are self-explanatory, like improving performance and fixing linting. The others are simply for consistency - I see ES6 is being used _most_ places right now but not all, so I changed some code so it also uses ES6 syntax. 

Commit c336853 is about consistency and personal preference for readability. I see you're using _that_ syntax in some other places and to me it's more readable, so I thought I'd change it. If you do not agree with some changes feel free to change, but in that case please let me know why. Thanks.

PR #207, opened 2017-01-26
Fix slider value calculation

lavrton:
How to reproduce:
1. Start dragging of slider
2. Move mouse to the left of the slider (or above the slider in a case of the vertical slider).
3. You will see that the slider value will be bigger than min value (and it starts moving in the right direction).

I guess it is unexpected behaviour.

### DIFF
--- a/src/createSlider.jsx
+++ b/src/createSlider.jsx
@@ -154,7 +154,7 @@ export default function createSlider(Component) {
 
     calcValue(offset) {
       const { vertical, min, max } = this.props;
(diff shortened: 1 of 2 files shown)


benjycui:
THX, it will be better if you can add a unit test in https://github.com/react-component/slider/blob/master/tests/common.test.js

lavrton:
@benjycui I added a test. It fails without the patch and it pass when the patch is applied.

benjycui:
+ rc-slider@6.0.1

PR #210, opened 2017-01-30
fix: ensure ref is present before using

STRML:
It's possible for the ref to be undefined, especially when using async batching strategies.

For example:

```
TypeError: Cannot read property 'getBoundingClientRect' of null
  at getSliderStart(node_modules/rc-slider/lib/createSlider.js:170:0)
  at calcValueByPos(node_modules/rc-slider/lib/createSlider.js:196:0)
  at onMove(node_modules/rc-slider/lib/Slider.js:111:0)
  at callback(node_modules/rc-slider/lib/createSlider.js:120:0)
  at batchedUpdates(ui/utils/ReactRAFBatchingStrategy.js:29:13)
  at unstable_batchedUpdates(node_modules/react-dom/lib/ReactUpdates.js:97:0)
  at call(node_modules/rc-util/lib/Dom/addEventListener.js:21:0)
  at apply(node_modules/add-dom-event-listener/lib/index.js:17:0)
  at apply(node_modules/raven-js/src/raven.js:278:0)
```

### DIFF
--- a/src/createSlider.jsx
+++ b/src/createSlider.jsx
@@ -66,6 +66,11 @@ export default function createSlider(Component) {
       }
     }
 
+    componentWillUnmount() {
+      if (super.componentWillUnmount) super.componentWillUnmount();
+      this.removeDocumentEvents();
+    }
+
     onMouseDown = (e) => {
       if (e.button !== 0) { return; }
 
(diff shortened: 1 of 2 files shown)


benjycui:
Could you add a unit test for this in https://github.com/react-component/slider/blob/master/tests/common.test.js ? THX.

STRML:
Updated - also noticed that these event handlers could remain in the DOM if the component is unmounted before mouseup/touchend. Added a `componentWillUnmount` and corresponding test.

benjycui:
6.0.2

PR #214, opened 2017-02-03
Improves scale-handling, handle-marks and adds eslintrc

TN1ck:
Hey!

This is a really cool project, I needed some new features though, which I think would benefit more people:

## Scale handling
The current code is pretty hard on the assumption that the scale is always linear (to be seen at `calcOffset` and `calcValue`). I added support for custom scale-functions and my PR adds a `pow`-scale as an example. Currently I use `d3-scale` for this, but if wanted, I could manually implement some scales to have good set of defaults.
I also changed the name of the function `calcOffset` to `calcOffsetPercentage`, because it confused me with the parameter that is given to `calcValue`. I think this makes it more clear.

## Handle-marks
I added the ability to add marks to all handles. This is done via a function which takes the arguments of the value of the handle and its index.

## .eslintrc
I added a .eslintrc-config which should catch the spirit of the currently used style-guide. This makes it easier for future people to contribute and helps to define what style is used here.

So what do you guys think? Do you think these features benefit the project? Should I split them into multiple PRs? What needs to be changed that you'll include them (removal of the new package?).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,198 @@
+{
+    "parser": "babel-eslint",
+    // there is a issue with used-vars inside classes https://github.com/babel/babel-eslint/issues/21
+    "plugins": [
+        "react"
+    ],
+    "env": {
+        "browser": true,
+        "node": true
+    },
+    "globals": {
+        "Promise": false,
+        "Tracking": false
+    },
+    "ecmaFeatures": {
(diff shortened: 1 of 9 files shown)


benjycui:
Sorry, I want to keep `rc-slider` simple and flexible. And `Slider[scale]` is easy to implement with HOC.

Issue #46, opened 2015-11-20
High maximum value / step causes errors


iktl:
As the slider's max value increases performance degrades steadily, before failing outright at higher values. This looks to be a function of max value and step together (the number of actual steps on the slider). I was trying this out as a currency amount input slider so the max value was well into the tens / hundreds of thousands. 

The error is:

```
Unhandled promise rejection RangeError: Maximum call stack size exceeded(…)
```

I was getting this at: 
* `max = 200,000` and `step = 1`
* `max = 100,000` and `step = 0.5`

benjycui:
Yes, and I knew the reason for low performance. I will try to make it faster when I have time.

But I am curious about your use case. It is hard to select a specific point out of 200,000 points.

iktl:
It's definitely a weirder use case, where users are able to adjust the sliders and modify data points being pushed onto a chart – so the slider values are linked to the domain of the chart itself. However, that being said, I changed the approach to instead of using the raw value from the slider:

* Keep the min / max values as a prop on my container
* Set the slider from 0 - 100
* onChange function takes the slider's input from 0 to 100% and interpolates it between the min and the max.

The result ends up letting me user the slider to select from a very, very wide range, but without needing to create 100,000+ steps in the slider itself. The workaround is solid for me, so if you even wanted to just close this issue outright and if it comes up again suggest this as the workaround I'm fine with that.

PR #228, opened 2017-02-27
[WIP] fix: decimal step cause divide not equal

### LINKED ISSUE #227: Pushable only works in positive direction with decimal steps
Not sure if this is a proper bug really, since maybe it shouldn't be possible to use decimals? I'm reporting it in any case:

The following example should demonstrate an issue where if you drag the leftmost handle to the right you can push the other handles, but dragging the rightmost handle to the left does not let you push the others.

http://codepen.io/oddsor/pen/vxOLVO

I'm not sure if this bug is known or if it's something that is easily fixable, so I've just worked around it by ensuring that I use "integers".

paranoidjk:
close #227

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -168,25 +168,24 @@ class Range extends React.Component {
     return bounds[bounds.length - 1];
   }
 
-  /**
-   * Returns an array of possible slider points, taking into account both
-   * `marks` and `step`. The result is cached.
-   */
-  getPoints() {
-    const { marks, step, min, max } = this.props;
-    const cache = this._getPointsCache;
-    if (!cache || cache.marks !== marks || cache.step !== step) {
-      const pointsObject = { ...marks };
-      if (step !== null) {
-        for (let point = min; point <= max; point += step) {
-          pointsObject[point] = point;
-        }
-      }
-      const points = Object.keys(pointsObject).map(parseFloat);
-      points.sort((a, b) => a - b);
-      this._getPointsCache = { marks, step, points };
-    }
-    return this._getPointsCache.points;
+  pointsRule() {
+    const { step, min, max } = this.props;
+    const pointsLength = (max - min) / step;
+    const getPoint = (index) => {
+      return index * step;
+    };
+    const getIndex = (val) => {
+      return val >= min && val <= max && val / step;
+    };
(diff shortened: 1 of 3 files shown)


paranoidjk:
ping @benjycui

src/Range.jsx
```diff
@@ -176,10 +176,17 @@ class Range extends React.Component {
     const { marks, step, min, max } = this.props;
     const cache = this._getPointsCache;
     if (!cache || cache.marks !== marks || cache.step !== step) {
+      const decimal = step.toString().split('.')[1] && step.toString().split('.')[1].length;
```
benjycui: Use [`getPrecision`](https://github.com/react-component/slider/blob/4e26da531406305e1c2b9396d13c84c024193fba/src/utils.js#L28-L35) ？

src/Range.jsx
```diff
@@ -176,10 +176,17 @@ class Range extends React.Component {
...
       const pointsObject = { ...marks };
       if (step !== null) {
-        for (let point = min; point <= max; point += step) {
+        let point = min;
+        while (point <= max) {
```
benjycui: This loop will cost a lot.

e.g. min = 0 && max = 99999

src/Range.jsx
paranoidjk: ok

src/Range.jsx
benjycui: https://github.com/react-component/slider/issues/46

paranoidjk: the performance problem only occur when users set `ste` to be a very small decimals, I think it's not our worries ?

paranoidjk: maybe use something like `binary search` + `Array.map` to caculate the point more quickly?

paranoidjk: ~~caculate the point more quickly~~ ——> avoid Maximum call stack

paranoidjk: Another idea,  what if when `step` is very small compare to `max - min`, we don't calculate the whole points array at first, instead just get the formula，something like： `value = f(index)`, then the cost just happen when every time getter

paranoidjk:
close. going to open a new pr to fix the decimal step problem.

PR #232, opened 2017-03-03
fix: noticket - Re-add tipFormatter to Slider API

logicalicy:
## Summary

Would like to add the `tipFormatter` prop back to the `Slider` API, as in the [`Slider` API of old (v5.4.3)](https://github.com/react-component/slider/blob/f7aacbfe0a568878630b8a18eadc035fcf9c2937/src/Slider.jsx#L458). Is this the best way?

For example, if I want to format a number with a comma for every thousand:

```
const SliderWithTooltip = ReactSlider.createSliderWithTooltip(ReactSlider);

<SliderWithTooltip tipFormatter={tipFormatter} ... />
```

image

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -109,6 +109,7 @@ class Slider extends React.Component {
       vertical,
       included,
       disabled,
+      tipFormatter,
       handle: handleGenerator,
     } = this.props;
     const { value, dragging } = this.state;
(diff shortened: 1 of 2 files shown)


benjycui:
And you need to update README.md at the same time.

TincaTibo:
Hi,

Same changes works as well for Range API
It would be great to include them as well.

Thibaut

benjycui:
> You don't need to change src/slider.jsx
> 
> Just update src/createSliderWithTooltip.

....

benjycui:
I will do this later.

benjycui:
trace: https://github.com/react-component/slider/pull/236

PR #236, opened 2017-03-13
feat: add tipFormatter for Slider/Range with tooltip, close: #232

benjycui:
close: #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,12 @@ const Range = createSliderWithTooltip(Slider.Range);
 
 [Online demo](http://react-component.github.io/slider/examples/handle.html)
 
+After Range or Slider was wrapped by createSliderWithTooltip, it will have the following props:
+
+| Name         | Type    | Default | Description |
+| ------------ | ------- | ------- | ----------- |
+| tipFormatter | (value: number): React.ReactNode | value => value | A function to format tooltip's overlay |
+
 ### Common API
 
 The following APIs are shared by Slider and Range.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -80,6 +80,12 @@ const Range = createSliderWithTooltip(Slider.Range);
 
 [Online demo](http://react-component.github.io/slider/examples/handle.html)
 
+After Range or Slider was wrapped by createSliderWithTooltip, it will have the following props:
+
+| Name         | Type    | Default | Description |
+| ------------ | ------- | ------- | ----------- |
+| tipFormatter | (value: number): React.ReactNode | value => value | A function to format tooltip overlay, if it's null, tooltip will be invisible |
```
paranoidjk: >  if it's null, tooltip will be invisible

This logic seems kind of weird，If one use `createSliderWithTooltip`, I think he wants to show tooltip.

`tipFormatter` is  just a way of data transform which default value is not transform

benjycui: Yep, I just use prev logic...

benjycui:
Updated

paranoidjk:
👍 +1

benjycui:
@paranoidjk so, just merge it.. and publish

paranoidjk:
@benjycui Coverage decreased

benjycui:
😢 it's not easy to test thing like tooltip, just ignore it...

paranoidjk:
It's a little strange, you just add one single api, and Coverage decreased `(-4.5%)`, no need to worry why?

paranoidjk:
6.2.0

benjycui:
Possible reason: https://github.com/nickmerwin/node-coveralls/issues/158

PR #247, opened 2017-03-31
feat(train-slide): adds train slider

niksinfinite10:


benjycui:
What are you going to do?

I don't understand your use case.

PR #251, opened 2017-04-07
Add handleStyle and minimumTrackStyle props to Range

gragland:
Previously the `handleStyle` and `minimumTrackStyle` props only worked with Slider. Added them to Range as well.

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -281,6 +281,8 @@ class Range extends React.Component {
       vertical,
       included,
       disabled,
+      minimumTrackStyle,
+      handleStyle,
       handle: handleGenerator,
     } = this.props;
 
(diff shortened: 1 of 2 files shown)


paranoidjk:
since Range has multiple track and handle，so the api name may be confused, we need to think about it.

https://github.com/react-component/slider/pull/249

gragland:
Maybe innerTrackStyle and outerTrackStyle would be more appropriate names. It makes sense for Range, and for the slider I don't think it's too much of a stretch to consider the track to the left of the handle to be "inner".. What do you think?

paranoidjk:
@benjycui

gragland:
Opened a new pull request for this since this one was pointing to my master branch. Sorry for the confusion. See: https://github.com/react-component/slider/pull/255

PR #252, opened 2017-04-08
Migrated deprecated React.PropTypes and React.createClass

Andarist:


paranoidjk:
Add related info: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

paranoidjk:
👍  Thanks

benjycui:
@Andarist how about refactoring this to ES6 class?

Andarist:
@benjycui 
yeah, that's something I was thinking about too, but had not that much time for updates, I've kinda went for a run and helped few libraries with the migration at once, so Ive wanted to keep changes to the minimum

That is certainly something that should be prepared before `v16`, so by then it's already ready. You can create an issue about the matter and mention me there - if I find time Im gonna do it, but no hard promises though.

PR #253, opened 2017-04-10
Refactored examples to use es6 classes

Andarist:


examples/range.js
```diff
@@ -13,30 +12,31 @@ function log(value) {
...
+  constructor(props) {
+    super(props);
+    this.state = {
       lowerBound: 20,
       upperBound: 40,
       value: [20, 40],
     };
-  },
+  }
   onLowerBoundChange(e) {
```
paranoidjk: prefer to use arrow function, in case of `this` problem.

``` js
   onLowerBoundChange = (e) {
    }

```

examples/range.js
```diff
@@ -13,30 +12,31 @@ function log(value) {
...
+class CustomizedRange extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
       lowerBound: 20,
       upperBound: 40,
       value: [20, 40],
     };
-  },
+  }
```
paranoidjk: prefer to use arrow function, in case of `this` problem.

``` js
   onLowerBoundChange = (e) {
    }

```

examples/range.js
paranoidjk: prefer to use arrow function, in case of `this` problem.

``` js
   onLowerBoundChange = (e) {
    }

```

Andarist: oh, sure thing, didnt notice this out of a hurry

fixed!

PR #255, opened 2017-04-13
Add handleStyle and minimumTrackStyle props to Range

gragland:
Previously the `handleStyle` and `minimumTrackStyle` props only worked with Slider. Added them to Range as well.

Maybe they could be renamed to `innerTrackStyle` and `outerTrackStyle`. It makes sense for Range, and for the slider I don't think it's too much of a stretch to consider the track to the left of the handle to be "inner".

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -282,6 +282,8 @@ class Range extends React.Component {
       vertical,
       included,
       disabled,
+      minimumTrackStyle,
+      handleStyle,
       handle: handleGenerator,
     } = this.props;
 
@@ -299,6 +301,7 @@ class Range extends React.Component {
       dragging: handle === i,
       index: i,
       disabled,
+      handleStyle,
       ref: h => this.saveHandle(i, h),
     }));
 
@@ -315,6 +318,7 @@ class Range extends React.Component {
           included={included}
           offset={offsets[i - 1]}
           length={offsets[i] - offsets[i - 1]}
+          minimumTrackStyle={minimumTrackStyle}
           key={i}
         />
       );
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -18,6 +18,14 @@ class Range extends React.Component {
     ]),
     allowCross: PropTypes.bool,
     disabled: PropTypes.bool,
+    rangeTrackStyle: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.object),
+      PropTypes.object,
+    ]),
+    rangeHandleStyle: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.object),
+      PropTypes.object,
+    ]),
   };
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


paranoidjk:
what about this,  if a Array is been provided, then we use it as style of track/handle follow the array index order, otherwise if A single style object is been provided, use it for all track/handle

```js
rangeHandleStyle: Array[Style Object] | Style Object
rangeTrackStyle: Array[Style Object] | Style Object
```

gragland:
@paranoidjk That sounds great to me!

paranoidjk:
@gragland would you like PR for this ? sorry that I recently have no time to do this feature

gragland:
@paranoidjk Sure I can do a PR for this within the next couple days.

src/Range.jsx
```diff
@@ -282,6 +282,8 @@ class Range extends React.Component {
       vertical,
       included,
       disabled,
+      minimumTrackStyle,
```
benjycui: Just `trackStyle`

And you should update README.md at the same time.

benjycui:
@paranoidjk just use a semantic name that you prefer.

senkevichdv:
Any news with this feature?

mergebandit:
Would love to see this merged!

paranoidjk:
@gragland please add api doc on readme and demo on examples at the same time.

benjycui:
Oops conflict...

paranoidjk:
@billycai  You can leave your comments at here, let us konw what do you think about the api.

https://github.com/react-component/slider/issues/277

donaldtf:
Any idea when this might be merged in?

chongjunhong:
Hi guys, has the feature been pushed to production? Thanks :)

paranoidjk:
Thanks for @gragland ,  I will handle this, the api will looks like this:

## `rc-slider@8.x`, which is used for `antd@2.x`, `antd-mobile@1.x`

| Name         | Type    | Default | Description |
| ------------ | ------- | ------- | ----------- |
| minimumTrackStyle | Object |`{}` | The style used for the track to the left of the button. (`only used for slider, just for compatibility , will be deprecate `) |
| maximumTrackStyle | Object | `{}` | The style used for the track to the right of the button. (`only used for slider, just for compatibility , will be deprecate `) |
| handleStyle | Array[Object] \| Object | `[{}]` | The style used for handle. (`both for slider and range, the array will be used for mutli handle follow element order`) |
| trackStyle | Array[Object] \| Object | `[{}]` | The style used for track.  . (`both for slider and range, the array will be used for mutli track follow element order`)|

## `rc-slider@9.x`, which is used for `antd@3.x`, `antd-mobile@2.x`

| Name         | Type    | Default | Description |
| ------------ | ------- | ------- | ----------- |
| handleStyle | Array[Object] \| Object | `[{}]` | The style used for handle. (`both for slider and range, the array will be used for mutli handle`) |
| trackStyle | Array[Object] \| Object | `[{}]` | The style used for track.  . (`both for slider and range, the array will be used for mutli track`)|

cc @benjycui

paranoidjk:
I opened a PR,  look forward to you guys review https://github.com/react-component/slider/pull/281

PR #272, opened 2017-05-19
Preparing node module

unsigned6:
* Prepared for modifying and including as node module.
* Complied and add missing files needed for nodule.
* to add to your project in package.json add next:
`"rc-slider": "https://github.com/react-component/slider.git"`
* after each change in `src/` need execute `npm run compile` and `git add .` to include all needed in build files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,5 @@ node_modules
 dist
 assets/**/*.css
 build
-lib
-es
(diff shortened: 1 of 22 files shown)


paranoidjk:
these files do not need to add in git， https://unpkg.com/rc-slider@7.0.8/

PR #278, opened 2017-06-01
feat: compatible with 7.x, in case uers do not pass value to handle.

paranoidjk:


paranoidjk:
cc @silentcloud 之前是我理解错了，slider 做一下兼容，没有 break change 了。

benjycui:
👍

PR #280, opened 2017-06-07
add tipProps on tooltipslider

### LINKED ISSUE #276: rc-tooltip in a react bootstrap Modal
Hi, 
I got a issue when using the rc-tooltip in a react bootstrap Modal whereby the tooltip is behind the modal and therefore not visible.
Any idea on how to fix that?

zeakd:
After createSliderWIthTooltip, slider can get tipProps.  close #276

### DIFF
--- a/src/createSliderWithTooltip.jsx
+++ b/src/createSliderWithTooltip.jsx
@@ -8,6 +8,7 @@ export default function createSliderWithTooltip(Component) {
     static propTypes = {
       tipFormatter: PropTypes.func,
       handleStyle: PropTypes.arrayOf(PropTypes.object),
+      tipProps: PropTypes.object,
     };
     static defaultProps = {
       tipFormatter(value) { return value; },
@@ -28,14 +29,27 @@ export default function createSliderWithTooltip(Component) {
       });
     }
     handleWithTooltip = ({ value, dragging, index, disabled, ...restProps }) => {
-      const { tipFormatter, handleStyle } = this.props;
+      const {
+        tipFormatter,
+        tipProps,
+        handleStyle,
+      } = this.props;
+
+      const {
+        prefixCls = 'rc-slider-tooltip',
+        overlay = tipFormatter(value),
+        placement = 'top',
+        ...restTooltipProps,
+      } = tipProps;
+
       return (
         <Tooltip
-          prefixCls="rc-slider-tooltip"
-          overlay={tipFormatter(value)}
           visible={!disabled && (this.state.visibles[index] || dragging)}
-          placement="top"
           key={index}
+          prefixCls={prefixCls}
+          overlay={overlay}
+          placement={placement}
+          {...restTooltipProps}
         >
           <Handle
             {...restProps}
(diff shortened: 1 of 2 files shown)


src/createSliderWithTooltip.jsx
```diff
@@ -26,14 +28,26 @@ export default function createSliderWithTooltip(Component) {
...
         <Tooltip
-          prefixCls="rc-slider-tooltip"
-          overlay={tipFormatter(value)}
           visible={!disabled && (this.state.visibles[index] || dragging)}
-          placement="top"
           key={index}
+          prefixCls={prefixCls}
+          overlay={overlay}
+          placement={placement}
+          {...restTooltipProps}
```
benjycui: `{...restTooltipProps}` should be the first props.

zeakd: is there any reason?

benjycui: I afraid that user will override `visible`

benjycui: This may cause some issues, if users are not familiar with implementation of rc-slider.

zeakd: you are right. How about just destruct visible from tipProps? IMO ordering sensitive props makes it hard to maintain.

paranoidjk: @zeakd actually, many of our code rely on props ordering to control the props override logic

zeakd: I see. I updated it

src/createSliderWithTooltip.jsx
```diff
@@ -26,13 +28,25 @@ export default function createSliderWithTooltip(Component) {
...
-      const { tipFormatter } = this.props;
+      const {
+        tipFormatter,
+        tipProps,
+      } = this.props;
+
+      const {
+        prefixCls = 'rc-slider-tooltip',
+        overlay = tipFormatter(value),
+        placement = 'top',
```
paranoidjk: I think `placement` should be one of restTooltipProps, as for the defaultValue, just use:

```tsx
placement="top"
{...restTooltipProps}
overlay={tipFormatter(value)}	
prefixCls={prefixCls}
overlay={overlay}
```

zeakd: Sorry, I have two question.
1. what is difference? `placement` is already have default value. is it convention?
2. how about `prefixCls`? I think `prefixCls` should have same form with `placement` in your suggestion. like 
```
placement='top'
prefixCls='rc-slider-tooltip'
{...restTooltipProps}
overlay={tipFormatter(value)}
overlay={overlay} 
visible={!disabled && (this.state.visibles[index] || dragging)}
key={index}
```

paranoidjk: @zeak 

The difference is maintainability. But i think right now is ok. I will merge this and release a minor version. Thanks.

paranoidjk:
@zeakd  can you rebase master and resolve the conflict?

zeakd:
Oh there is new feature `handleStyle`. I'll do it

Issue #131, opened 2016-08-05
Not accessible


jzaefferer:
My project team is considering adopting this component. After a brief review I noticed that the component isn't accessible at all: For example, there are no focusable elements, and accordingly, no way to interact with the keyboard. It apparently handles both mouse and touch events, but ignores pointer events.

Making the handle focusable is rather trivial, by adding `tabindex=0` (see [jQuery UI slider widget](https://github.com/jquery/jquery-ui/blob/3dd8a09b441d65445f2b6a7c73e72af65445d5da/ui/widgets/slider.js#L99)). Handling keyboard events isn't terrible either (see [jQuery UI slider widget, _handleEvents](https://github.com/jquery/jquery-ui/blob/3dd8a09b441d65445f2b6a7c73e72af65445d5da/ui/widgets/slider.js#L670-L748)).  Switching from mouse+touch events to pointer events would require a polyfill like [PEP](https://github.com/jquery/PEP) but should actually simplify the code inside the component. There's probably more that I'm missing.

That said, is there interest addressing this?

benjycui:
Yep, you are right. I am not sure whether you are interested in creating a PR, or I will do it when I have time..

jzaefferer:
Sorry, can't help with the implementation currently.

slavikdenis:
+1

jbccollins:
@jzaefferer I encountered the same issue. My use case was that I wanted to use a custom thumb handle and apply styles like a border highlight and z-index to handles that were being dragged. The ":active" and ":focus" CSS selectors were not working (except in Chrome). However, I was able to apply styles by checking the props of the handle class. There is a boolean "dragging" prop which is only true when a handle is currently being dragged. Here's an example:
```javascript
<Slider
  {...sliderProps}
  handle={<CustomHandle />}
/>
...
class CustomHandle extends React.Component {
  render() {
    const {
      value,
      dragging, 
    } = this.props;

    const activeClass = dragging ? "active" : "inactive";
    const style = { left: `${this.props.offset}%` };
    return (
      <div style={style} className={"drag-handle " + activeClass}>
          value
      </div>
    );
  }
}
```

Hope that helps.

kylecesmat:
I'd be interested i helping to make this component accessible. Currently, the API does not allow for passing callbacks to custom Handle components (from what I can tell) meaning that `CustomHandle` implementations can't do much aside from modify style.

[...]

PR #282, opened 2017-06-15
Basic slider accessibility

byCedric:
As promptly discussed in issue #131, I created some basic accessibility for the Slider. I implemented the keyboard behaviour as my personal best guess of how it should work. 

> Note, this is a work in progress but I wanted to share my changes already.

### DIFF
--- a/src/Slider.jsx
+++ b/src/Slider.jsx
@@ -94,6 +94,21 @@ class Slider extends React.Component {
     this.onChange({ value });
   }
 
+  onKeyboard(e) {
+    const valueMutator = utils.getKeyboardValueMutator(e);
+
+    if (valueMutator) {
+      utils.pauseEvent(e);
+      const state = this.state;
+      const oldValue = state.value;
+      const mutatedValue = valueMutator(oldValue, this.props);
+      const value = this.trimAlignValue(mutatedValue);
+      if (value === oldValue) return;
+
+      this.onChange({ value });
+    }
+  }
+
   getValue() {
     return this.state.value;
   }
--- a/src/common/createSlider.jsx
+++ b/src/common/createSlider.jsx
@@ -120,6 +120,22 @@ export default function createSlider(Component) {
       utils.pauseEvent(e);
     }
 
+    onFocus = (e) => {
+      const isVertical = this.props.vertical;
+
+      if (utils.isEventFromHandle(e, this.handlesRefs)) {
+        const handlePosition = utils.getHandleCenterPosition(isVertical, e.target);
+
+        this.dragOffset = 0;
+        this.onStart(handlePosition);
+        utils.pauseEvent(e);
+      }
+    }
+
+    onBlur = (e) => {
+      this.onEnd(e);
+    };
+
     addDocumentTouchEvents() {
       // just work for Chrome iOS Safari and Android Browser
       this.onTouchMoveListener = addEventListener(document, 'touchmove', this.onTouchMove);
@@ -158,6 +174,12 @@ export default function createSlider(Component) {
 
       const position = utils.getTouchPosition(this.props.vertical, e);
       this.onMove(e, position - this.dragOffset);
+    }
+
+    onKeyDown = (e) => {
+      if (this.sliderRef && utils.isEventFromHandle(e, this.handlesRefs)) {
+        this.onKeyboard(e);
+      }
     }
 
     getSliderStart() {
@@ -237,6 +259,9 @@ export default function createSlider(Component) {
           className={sliderClassName}
           onTouchStart={disabled ? noop : this.onTouchStart}
           onMouseDown={disabled ? noop : this.onMouseDown}
+          onKeyDown={disabled ? noop : this.onKeyDown}
+          onFocus={disabled ? noop : this.onFocus}
+          onBlur={disabled ? noop : this.onBlur}
           style={style}
         >
           <div
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,5 @@
 import { findDOMNode } from 'react-dom';
+import keyCode from 'rc-util/lib/KeyCode';
 
 export function isEventFromHandle(e, handles) {
   return Object.keys(handles)
@@ -69,4 +70,23 @@ export function ensureValuePrecision(val, props) {
 export function pauseEvent(e) {
   e.stopPropagation();
   e.preventDefault();
+}
+
+export function getKeyboardValueMutator(e) {
+  switch (e.keyCode) {
+    case keyCode.UP:
+    case keyCode.RIGHT:
+      return (value, props) => value + props.step;
+
+    case keyCode.DOWN:
+    case keyCode.LEFT:
+      return (value, props) => value - props.step;
+
+    case keyCode.END: return (value, props) => props.max;
+    case keyCode.HOME: return (value, props) => props.min;
+    case keyCode.PAGE_UP: return (value, props) => value + props.step * 2;
+    case keyCode.PAGE_DOWN: return (value, props) => value - props.step * 2;
+
+    default: return undefined;
+  }
 }
--- a/assets/index.less
+++ b/assets/index.less
@@ -69,6 +69,11 @@
       cursor: -webkit-grabbing;
       cursor: grabbing;
     }
+    &:focus {
+      border-color: tint(@primary-color, 20%);
+      box-shadow: 0 0 0 5px tint(@primary-color, 50%);
+      outline: none;
+    }
   }
 
   &-mark {
--- a/src/Handle.jsx
+++ b/src/Handle.jsx
@@ -25,6 +25,7 @@ export default class Handle extends React.Component {
     return (
       <div
         role="slider"
+        tabIndex="0"
         {...ariaProps}
         {...restProps}
         className={className}
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import shallowEqual from 'shallowequal';
+import warning from 'warning';
 import Track from './common/Track';
 import createSlider from './common/createSlider';
 import * as utils from './utils';
@@ -134,6 +135,10 @@ class Range extends React.Component {
       handle: nextHandle,
       bounds: nextBounds,
     });
+  }
+
+  onKeyboard() {
+    warning(true, 'Keyboard support is not yet supported for ranges.');
   }
 
   getValue() {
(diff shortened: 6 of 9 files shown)


benjycui:
Oops, CI failed

paranoidjk:
@byCedric 
Thanks for your pr, Wonderful job 👍 
Could you please rebase master to solve the conflict and fix the ci, then i will do a code review.

byCedric:
@paranoidjk thanks! Yeah, of course, I have some time late tonight. I think I can have this done tomorrow morning, around 08:00 (GMT) or something like that.

byCedric:
@paranoidjk Ok, so I rebased and added the tabindex to the handlers within the snapshots. Can I help with something for the final merge?

benjycui:
👍 

It will be better if you can add unit test for this new feature.

src/utils.js
```diff
@@ -70,3 +71,22 @@ export function pauseEvent(e) {
...
+    case keyCode.RIGHT:
+      return (value, props) => value + props.step;
+
+    case keyCode.DOWN:
+    case keyCode.LEFT:
+      return (value, props) => value - props.step;
+
+    case keyCode.END: return (value, props) => props.max;
+    case keyCode.HOME: return (value, props) => props.min;
+    case keyCode.PAGE_UP: return (value, props) => value + props.step * 2;
```
paranoidjk: why `* 2` ？

as you described in https://github.com/react-component/slider/issues/131#issuecomment-308567343, peg_up should be equal with up

src/Slider.jsx
```diff
@@ -94,6 +94,21 @@ class Slider extends React.Component {
     this.onChange({ value });
   }
 
+  onKeyboard(e) {
+    const valueMutator = utils.getKeyboardValueMutator(e);
+
+    if (valueMutator) {
+      utils.pauseEvent(e);
+      const state = this.state;
+      const oldValue = state.value;
```
paranoidjk: no need to define a variable `state` ?

```js
const { value: oldValue } = this.state;
```

src/common/createSlider.jsx
```diff
@@ -120,6 +120,22 @@ export default function createSlider(Component) {
       utils.pauseEvent(e);
     }
 
+    onFocus = (e) => {
+      const isVertical = this.props.vertical;
+
+      if (utils.isEventFromHandle(e, this.handlesRefs)) {
+        const handlePosition = utils.getHandleCenterPosition(isVertical, e.target);
+
+        this.dragOffset = 0;
```
paranoidjk: Why set `dragOffset ` to zero?

src/utils.js
byCedric: > For the up and down behaviours I tend to follow the Polymer approach. Mainly because you are "upping" or "downing" the current value.

What I tried to explain was that I like how Polymer's approach had the same type of behaviour with both `up` and `pageup` keys. Both "up" keys increased the current value. OAA's behaviour was the same for `up` and `pagedown`, which I didn't like because of mixed "up" and "down" keys.

Later on I tried to reason that OAA has a point about the increased value jump through `pageup` and `pagedown`. For small sliders it makes no different, but if you have a big one you would probably want to go faster through the steps. 

> OAA Accessibility has a point about the page up and page down behaviours. I mean, left/down and up/right only go by 1 step. For small sliders this is fine, but for large ones not. Also if an user actually uses the page up/page down key, they probably expect something more then just a mimick of left and right.

I'm open for both approaches. Either increase the step when using the page keys or just copy the same behaviour from the normal up/down keys. It's up to you guys.

src/Slider.jsx
byCedric: If it was up to me, I would have definitely used these destructuring assignment. But I didn't want to introduce a new code styling. That's why I sticked as close as possible to the existing one. I think I based this part of the code on [this code](https://github.com/react-component/slider/blob/master/src/Range.jsx#L82-L106).

src/common/createSlider.jsx
byCedric: I think I did this to "reset" mouse/touch behaviours when combining keyboard with these. But I'm not sure if this is even necessary. Just tested with and without this and I had the same outcome... I can remove it if you want?

src/utils.js
paranoidjk: Ok, i understand. But please make sure you update the api doc in readme correctly.

src/Slider.jsx
paranoidjk: 😅 no need to obey current  code style, just use what you think is better.

It's just because that we do not have enough time to refactor this component right now, and PR is welcome ~

src/common/createSlider.jsx
paranoidjk: I think we can keep it, since  `this.dragOffset = 0;` is to restore drag status.

paranoidjk:
@byCedric 
We almost done. Please update the api doc on readme, and add some test case, you can take a look at the existing testcase in `/tests` folder.

src/Slider.jsx
byCedric: Haha, good to know! Let's keep this for now unchanged. If I have some time again after this PR, I will try to create one for some refactoring 😄

byCedric:
Yeah, sure! I will try to start on that tonight!

benjycui:
ping~ @byCedric

byCedric:
Hey guys, sorry it took a while to respond. I tried to write some tests for the slider first, but soon encountered some errors. For example, this is the most simple test I've written.

```javascript
  it('increments the value when key "up" was pressed', () => {
    const wrapper = mount(<Slider value={50} step={1} />);
    const handler = wrapper.find('.rc-slider-handle');

    handler.simulate('focus');
    handler.simulate('keyDown', { keyCode: keyCode.UP });

    expect(wrapper.state('value')).toBe(51);
  });
```

I think this should work right? But It gives me this error.

``` bash
$ npm test

> rc-slider@8.1.5 test /absolute/path/to/slider
> jest

 FAIL  tests/Slider.test.js
  ● Slider › increments the value when up key was pressed

    TypeError: Cannot read property 'toFixed' of undefined
      
      at Object.ensureValuePrecision (src/utils.js:75:2549)
      at ComponentEnhancer.trimAlignValue (src/Slider.jsx:127:20)
      at ComponentEnhancer.calcValueByPos (src/common/createSlider.jsx:224:1870)
      at ComponentEnhancer.onStart (src/Slider.jsx:85:1605)
      at ComponentEnhancer._this.onFocus (src/common/createSlider.jsx:130:45)
      at Object.invokeGuardedCallback [as invokeGuardedCallbackWithCatch] (node_modules/react-dom/lib/ReactErrorUtils.js:26:5)
      at executeDispatch (node_modules/react-dom/lib/EventPluginUtils.js:83:21)
      at Object.executeDispatchesInOrder (node_modules/react-dom/lib/EventPluginUtils.js:108:5)
      at executeDispatchesAndRelease (node_modules/react-dom/lib/EventPluginHub.js:43:22)
      at executeDispatchesAndReleaseSimulated (node_modules/react-dom/lib/EventPluginHub.js:51:10)
      at forEachAccumulated (node_modules/react-dom/lib/forEachAccumulated.js:26:8)
      at Object.processEventQueue (node_modules/react-dom/lib/EventPluginHub.js:252:7)
      at node_modules/react-dom/lib/ReactTestUtils.js:351:22
      at ReactDefaultBatchingStrategyTransaction.perform (node_modules/react-dom/lib/Transaction.js:143:20)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactDefaultBatchingStrategy.js:62:26)
      at Object.batchedUpdates (node_modules/react-dom/lib/ReactUpdates.js:97:27)
      at node_modules/react-dom/lib/ReactTestUtils.js:349:18
      at ReactWrapper.<anonymous> (node_modules/enzyme/build/ReactWrapper.js:776:11)
      at ReactWrapper.single (node_modules/enzyme/build/ReactWrapper.js:1421:25)
      at ReactWrapper.simulate (node_modules/enzyme/build/ReactWrapper.js:769:14)
      at Object.<anonymous> (tests/Slider.test.js:29:13)
      at handle (node_modules/worker-farm/lib/child/index.js:41:8)
      at process.<anonymous> (node_modules/worker-farm/lib/child/index.js:47:3)
      at emitTwo (events.js:125:13)
      at process.emit (events.js:213:7)
      at emit (internal/child_process.js:768:12)
      at _combinedTickCallback (internal/process/next_tick.js:141:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)

 PASS  tests/common/createSlider.test.js
 PASS  tests/common/marks.test.js
 PASS  tests/Range.test.js

Test Suites: 1 failed, 3 passed, 4 total
Tests:       1 failed, 1 skipped, 22 passed, 24 total
Snapshots:   3 passed, 3 total
Time:        5.482s
Ran all test suites.
npm ERR! Test failed.  See above for more details.
```

Any idea how to fix this? Or do you need actual code for this? @benjycui

benjycui:
@byCedric can you re-produce this with `master` branch?

paranoidjk:
@byCedric 

Please push your test code to this pr,  i'd like to help you figure out the problem.

And please make sure you do have rebased `master`.

byCedric:
Yeah, will do tonight. I'll start with the tests for the slider. It looks like I'm doing something wrong with trimming the value for keyboard events. But as I said, will rebase and add the slider tests tonight! 👍

byCedric:
Ok guys, I did a final effort in debugging what goes wrong. I found out that when focusing the handler directly, as coded in the example, the slider thinks he is exactly `0px` wide. The strange thing is that when I focus the wrapper first, and then fire a KeyboardEvent on the handler, everything works. This looks like it needs some extra caring and love, but for now I'm not going in to deep.

Right now I'm patching up the other tests, I'll commit it soon afterwards!

byCedric:
So what do you guys think about this, any tips/tricks I should do? I think there could be some improvements within the tests by e.g. grouping them. Or even automate some repeating tests by iterations, instead of hardcoding it. Anyways, I'm still a bit puzzled why I need to both use the `wrapper` _and_ `handler` for focus and keyboard event. Maybe you can shine your lights on that one @paranoidjk?

byCedric:
Oh wait, I see I changed a line I shouldn't have.

paranoidjk:
Thanks for @byCedric 👍  Released `rc-slider@8.3.0`

benjycui:
@paranoidjk PR to antd is welcomed.

paranoidjk:
@benjycui What do you mean? antd should be no need to do any change, right?

benjycui:
https://github.com/ant-design/ant-design/blob/master/package.json#L64

antd had locked minor

Issue #286, opened 2017-06-19
Clicking outside mark moves handle


jeznag:
![issue with mark](https://user-images.githubusercontent.com/1034890/27273787-efd5578c-5513-11e7-98c0-c074121b94b7.gif)

The mark shouldn't be 90% width. I'll try and come up with another way of positioning it. We just need to set the `left`

PR #287, opened 2017-06-19
fix($marks): Fix issue with width of mark

jeznag:
Mark width was too wide (90%) so if you clicked outside the mark, it moved the handle

#286

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ The following APIs are shared by Slider and Range.
 | min | number | `0` | The minimum value of the slider |
 | max | number | `100` | The maximum value of the slider |
 | marks | `{number: string}` or`{number: { style, label }}` | `{}` | Marks on the slider. The key determines the position, and the value determines what will show. If you want to set the style of a specific mark point, the value should be an object which contains `style` and `label` properties. |
+| markFontSize | number | 10 | The font size used for the mark labels. This is used to calculate the width of the mark labels. |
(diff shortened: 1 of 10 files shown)


benjycui:
Could you post a screenshot here?

assets/index.less
```diff
@@ -80,6 +80,7 @@
   }
 
   &-mark-text {
+    margin-left: -10px;
```
paranoidjk: text length is variable

jeznag:
image
Screenshot of fixed bug. I'm not sure if there's a better way to calculate the width of the text.

README.md
```diff
@@ -97,6 +97,7 @@ The following APIs are shared by Slider and Range.
 | min | number | `0` | The minimum value of the slider |
 | max | number | `100` | The maximum value of the slider |
 | marks | `{number: string}` or`{number: { style, label }}` | `{}` | Marks on the slider. The key determines the position, and the value determines what will show. If you want to set the style of a specific mark point, the value should be an object which contains `style` and `label` properties. |
+| markFontSize | number | 10 | The font size used for the mark labels. This is used to calculate the width of the mark labels. |
```
benjycui: I don't think this is a good idea, the actual font size is out of control.

jeznag: I don't know how else we can accurately calculate the width of a mark. Any suggestions?

benjycui:
👎 for this, actually, we can solve this issue, but it is expensive(need to re-calculate marks' width every time). And this issue is not so troublesome.

PR #292, opened 2017-07-01
Added dotStyle and activeDotStyle

tomoberry:
I was thinking something like this for adding styling to the dots


benjycui:
😒

paranoidjk:
@tomoberry 

Thanks for you pr. 
But since we have `rc-slider-dot`, `rc-slider-dot-active` classname, why not custom style with css?

tomoberry:
I have a use case where I need to support an arbitrary number of different color combinations for the slider.  A simple example would be making an ajax call to a server that returns a random color that the slider should be displayed in.

paranoidjk:
hmm, i got this reature request from another user too.

So i am going to merge this and release a minor upgrade.

paranoidjk:
`rc-slider@8.2.0`

PR #289, opened 2017-06-21
Fix #279 - Use getBoundingClientRect for proper scaled width of slider

dbryand:
Jcqt-390's solution also worked for me:
https://github.com/react-component/slider/issues/279

PR #293, opened 2017-07-01
Support sliders with CSS scale()

### LINKED ISSUE #279: Wrong handle position when scaled
The slider handle is not placed at the right position when it is scaled.

Step to reproduce : Place the slider inside a scaled div

const wrapperStyle = { width: 400, margin: 50,transform:scale(0.5) };
ReactDOM.render(
  <div>
    <div style={wrapperStyle}>
      <p>Slider with custom handle</p>
      <Slider min={0} max={20} defaultValue={3} handle={handle} />
    </div>
    <div style={wrapperStyle}>
      <p>Range with custom handle</p>
      <Range min={0} max={20} defaultValue={[3, 10]} tipFormatter={value => `${value}%`} />
    </div>
  </div>,
  document.getElementById('__react-content')
);

I changed line 211 in lib/common/CreateSlider.js and it works for me :

Prev :  return this.props.vertical ? slider.clientHeight : slider.clientWidth;
Now :  return this.props.vertical ? slider.getBoundingClientRect().height : slider.getBoundingClientRect().width;

captbaritone:
Fixes #279. Includes #289 with a small refactor and also resolves the failing tests.

### DIFF
--- a/tests/common/createSlider.test.js
+++ b/tests/common/createSlider.test.js
@@ -4,16 +4,22 @@ import { mount } from 'enzyme';
 import Slider from '../..';
 const { Range } = Slider;
 
-// https://github.com/tmpvar/jsdom/commit/0cdb2efcc69b6672dc2928644fc0172df5521176
-const polyfillJsDomApi = property => (object, value) => {
-  Object.defineProperty(object, property, {
-    get() { return value; },
+const setWidth = (object, width) => {
+  // https://github.com/tmpvar/jsdom/commit/0cdb2efcc69b6672dc2928644fc0172df5521176
+  Object.defineProperty(object, 'getBoundingClientRect', {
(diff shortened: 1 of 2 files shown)


tests/common/createSlider.test.js
```diff
@@ -97,9 +103,9 @@ describe('createSlider', () => {
   });
 
   // TODO: should update the following test cases for it should test API instead implementation
-  it('should set `dragOffset` to correct value when the left handle is clicked off-center', () => {
+  fit('should set `dragOffset` to correct value when the left handle is clicked off-center', () => {
```
benjycui: `fit` => `it`

captbaritone: Doh. Fixed.

PR #300, opened 2017-08-04
[WIP] Range keyboard implementation

byCedric:
As announced in PR #298 and build for PR #131 (accessibility) here is the implementation for the Range components. I kept this from the other PR to make sure everything was correctly implemented for the Slider first.

> This is a Work In Progress, right now it lacks tests.

Issue #298, opened 2017-07-30
Wrong handler is being focused when multiple handles cross


byCedric:
I was working on implementing keyboard support for the Range sliders as well. But as previously stated in the issue #131, I encountered some issues when using multiple handles within a Range slider. 

Take a look at this:

![s38lazuscn](https://user-images.githubusercontent.com/1203991/28753120-5fc6a326-752f-11e7-83a5-eeada8fde034.gif)

What I'm doing here is moving the handle, with the mouse, across the other handle. The tooltip is being rendered correctly, but unfortunately the "focus" is left on the wrong handle. Or, the wrong handle is being moved. For mouse control this isn't much big of a deal, but for keyboard control its really not that great.

![6ofduvfzna](https://user-images.githubusercontent.com/1203991/28753187-574b1dde-7530-11e7-9374-ffd05b08e0e8.gif)

paranoidjk:
@byCedric  Thanks for your report.

The logic is correct at here, when you move across handle, actually the first handle is stoped at the second handle start point, then what you are draging is actually the second handle.  

That is to say, when you move across handles, there is a active handle exchange happened, but users may not notice that,  they looks like the first hanlde is acrosss the second.

But there do have bug as you report, the active handle state did't exchange correctly.

Since this bug is not very crucially，and i am busy right now, i will handle this when i got time.

PS. PR is welcome ~

Robin-front:
and it seems doesn't support keyboardEvent to move handles. should it do ?

byCedric:
@Robin-front I was the one who implemented it for simple sliders in PR #282. Because of this issue I withheld the implementation for ranges. I will add the PR for keyboard support in ranges tonight (GMT +02:00), because this is now a separate issue.

@paranoidjk If you think its wise to not yet support keyboard for ranges because of this, you can decline the coming PR. 😉

nrps9909:
I revalidated this against current master `02260ea7a23e09a76f34d9c59d41c9aae8561140`, and the crossed-handle focus problem is no longer reproducible.

I added a temporary regression that:

1. rendered a range with `[20, 40]`;
2. focused the first real handle;
3. dragged it from 20 across the other handle to 60;
4. released the drag; and
5. asserted that `document.activeElement` was the second sorted handle with `aria-valuenow="60"`.

[...]

PR #302, opened 2017-08-08
dont exchange handle, just change the algorithm of track

Robin-front:
trace: #298 

about the wrong handle focus, I think we could try to keep the same handle instead of exchange handles. just change the algorithm of track for display correctly.

**but it will cause the handle reverse, no lowerhandle and upperhandle concept. it's potential risk.**

if you agree that, #300 should also make a change in `props.allowCross` logic.

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -123,13 +123,10 @@ class Range extends React.Component {
 
     const nextBounds = [...state.bounds];
     nextBounds[state.handle] = value;
-    let nextHandle = state.handle;
+    const nextHandle = state.handle;
     if (props.pushable !== false) {
       const originalValue = state.bounds[nextHandle];
       this.pushSurroundingHandles(nextBounds, nextHandle, originalValue);
-    } else if (props.allowCross) {
-      nextBounds.sort((a, b) => a - b);
-      nextHandle = nextBounds.indexOf(value);
     }
     this.onChange({
       handle: nextHandle,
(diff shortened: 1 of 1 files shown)


src/Range.jsx
```diff
@@ -123,13 +123,10 @@ class Range extends React.Component {
 
     const nextBounds = [...state.bounds];
     nextBounds[state.handle] = value;
-    let nextHandle = state.handle;
+    const nextHandle = state.handle;
     if (props.pushable !== false) {
       const originalValue = state.bounds[nextHandle];
       this.pushSurroundingHandles(nextBounds, nextHandle, originalValue);
-    } else if (props.allowCross) {
```
benjycui: You are removing a feature?!

Robin-front: I think no.
the codes about `allowCross` feature is here [https://github.com/react-component/slider/blob/master/src/Range.jsx#L270](https://github.com/react-component/slider/blob/master/src/Range.jsx#L270)
where I delete is update handle controller, isn't it ..

paranoidjk:
@RaoHai have you test with our example? Obviously there have some problem, the range value will be something like `[60, 50]` if lowerhandle across upperhandle.

Please read our dev doc https://github.com/react-component/slider#development and test more carefully. 😄

Robin-front:
@paranoidjk yes, I have to admit that there is a problem. That's why I said no lowerhandle and upperhandle concept.  If just the range output value is reverse, I think you can sort before output.

But, obviously, it's no the best resolvent. 
and maybe it will cause other problem.
e.g. order of focus is wrong because the dom order didn't change.
cross

PR #303, opened 2017-08-09
Update util.js

console0713:
if there is a scrollX and scrollX != 0, pageX will not changed, but getBoundingClientRect().left will change, the value will not correct,so i think the clientX is better


benjycui:
First, make sure you had found a bug. 

benjycui:
Close for no response.

Issue #159, opened 2016-10-07
Keep "pushable" between states/values


monsieurnebo:
Hi,

I'm storing the slider value in a redux state in the goal of keeping an history (you can switch between previous values by going back in your browser).

It seems that we can't keep the **pushable** props between two states using a different **value**.

If I don't specify any value in the Slider, the pushable props still work. If you set the value props (to update the display at each state), the pushable props stop working.

Any idea to fix this ?

benjycui:
@sosz could you review this issue?

patrickdevivo:
+1 I'm seeing similar issue, `pushable` prop only seems to work when `value` is not set

it4anna:
The same affect on allowCross={false} :(

dannypk:
Would be nice to have an overlapping = false or something, so when you have 2 values, they cannot be the same. I am trying to make a workaround, but seems that the values array mixes the values[0] with values[1] when they "overlap"

paranoidjk:
close since dulplicate, please trace https://github.com/react-component/slider/issues/347

Issue #226, opened 2017-02-20
Ranges with allowCross=false result in wrong handle moving when dragging


Oddsor:
Couldn't think of a better title to this

I encountered a strange issue when attempting to implement a sort of wrapper component around the rc-slider.

When implementing an onChange containing a setState, the third value in the range slider is set to the same value as the second when dragging the first. The setState does not necessarily need to modify values used by the rc-slider for this to break.

For instance: A range of [20, 66, 70] becomes [21, 66, 66] when dragging the first handle. This also fails when dragging the last handle.

I tracked this to the "ensureValueNotConflict"-function in Range.js component on line 336. It is called in the componentWillReceiveProps, and this is why it breaks:

    if (!allowCross && handle != null) {
      if (handle > 0 && val <= bounds[handle - 1]) {
        return bounds[handle - 1];
      }
      if (handle < bounds.length - 1 && val >= bounds[handle + 1]) {
        return bounds[handle + 1];
      }
    }

A hacky fix I came up with is to pass the index to ensureValueNotConflict and then use that to skip values that aren't next to the dragged value

    Range.prototype.componentWillReceiveProps = function componentWillReceiveProps(nextProps) {
        var _this2 = this;

        if (!('value' in nextProps || 'min' in nextProps || 'max' in nextProps)) return;
        var bounds = this.state.bounds;

        var value = nextProps.value || bounds;
        var nextBounds = value.map(function (v, i) { <----- Index!
          return _this2.trimAlignValue(v, i, nextProps);
        });

And then modifying ensureValueNotConflict like this:

    if (!allowCross && handle != null) {
      if (handle - 1 != index && handle + 1 != index){
        return val;
      }
      if (handle > 0 && val <= bounds[handle - 1]) {
        return bounds[handle - 1];
      }
      if (handle < bounds.length - 1 && val >= bounds[handle + 1]) {
        return bounds[handle + 1];
      }
    }

This seems to solve my issue

benjycui:
Could you provide re-producible online demo? And some steps to re-produce it. Thanks.

Oddsor:
I could put one up on codepen for instance, but I couldn't figure out how to add rc-slider as an external resource there.

paranoidjk:
use unpkg，something like https://unpkg.com/rc-slider@6.1.2

benjycui:
Please provide a re-producible demo: http://codepen.io/benjycui/pen/aJooRE?editors=0011

Oddsor:
[...]

PR #304, opened 2017-08-10
Bugfix/keep pushable

Robin-front:
trace: #159 
when check the values in `componentWillReceiveProps`  function, you use `map` but no index or Corresponding handle. The handle will be always what you are move.  
Then it invoke `ensureValueNotConflict ` function, but `ensureValueNotConflict ` just return prev or next handle value. (I think it should in consideration of `thershold`)

So the gap of pushable not work when [`allowCross=false`](https://github.com/react-component/slider/issues/159#issuecomment-263282912) and [set the value props (to update the display at each state)](https://github.com/react-component/slider/issues/159#issue-181678834)

I just do it. but maybe someone has more elegant code. Many places invoke `trimAlignValue ` func, so it may not `index` and `nextProps` args sometimes. 

base issue #159  , after fixed, I make some examples： 
all examples I set `pushable={10}` and `allowCross={false}`

pushable

pushable2

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -269,17 +269,16 @@ class Range extends React.Component {
 
   ensureValueNotConflict(i, val, { allowCross, pushable: thershold }) {
     const state = this.state || {};
-    const { handle, bounds } = state;
+    const { bounds } = state;
+    const handle = (i === undefined) ? state.handle : i;
     thershold = Number(thershold);
     /* eslint-disable eqeqeq */
-    if (!allowCross && handle != null) {
-      if (i === undefined || i === handle) {
-        if (handle > 0 && val <= (bounds[handle - 1] + thershold)) {
-          return bounds[handle - 1] + thershold;
-        }
-        if (handle < bounds.length - 1 && val >= (bounds[handle + 1] - thershold)) {
-          return bounds[handle + 1] - thershold;
-        }
+    if (!allowCross && handle != null && bounds !== undefined) {
+      if (handle > 0 && val <= (bounds[handle - 1] + thershold)) {
+        return bounds[handle - 1] + thershold;
+      }
+      if (handle < bounds.length - 1 && val >= (bounds[handle + 1] - thershold)) {
+        return bounds[handle + 1] - thershold;
       }
     }
     /* eslint-enable eqeqeq */
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -38,7 +38,7 @@ class Range extends React.Component {
             props.defaultValue : initialValue;
     const value = props.value !== undefined ?
             props.value : defaultValue;
-    const bounds = value.map(v => this.trimAlignValue(v));
+    const bounds = value.map((v, i) => this.trimAlignValue(v, i));
     const recent = bounds[0] === max ? 0 : bounds.length - 1;
 
     this.state = {
@@ -55,9 +55,10 @@ class Range extends React.Component {
         shallowEqual(this.props.value, nextProps.value)) {
       return;
     }
+
     const { bounds } = this.state;
     const value = nextProps.value || bounds;
-    const nextBounds = value.map(v => this.trimAlignValue(v, nextProps));
+    const nextBounds = value.map((v, i) => this.trimAlignValue(v, i, nextProps));
     if (nextBounds.length === bounds.length && nextBounds.every((v, i) => v === bounds[i])) return;
 
     this.setState({ bounds: nextBounds });
@@ -204,7 +205,6 @@ class Range extends React.Component {
   pushSurroundingHandles(bounds, handle, originalValue) {
     const { pushable: threshold } = this.props;
     const value = bounds[handle];
-
     let direction = 0;
     if (bounds[handle + 1] - value < threshold) {
       direction = +1; // push to right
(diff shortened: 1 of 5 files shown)


paranoidjk:
@Robin-front Could you add some test case about the problem you want to fix?

paranoidjk:
close https://github.com/react-component/slider/issues/347

paranoidjk:
@Robin-front Could you pleasse rebase master? I add a `Controlled Range, not allow across` demo in example, and run with you fix, still got problem when handle `i` is pushed to `i+1`, then any of other handle being not moveable.

src/Range.jsx
```diff
@@ -260,23 +260,25 @@ class Range extends React.Component {
...
+    const valNotConflict = this.ensureValueNotConflict(i, valInRange, mergedProps);
     return utils.ensureValuePrecision(valNotConflict, mergedProps);
   }
 
-  ensureValueNotConflict(val, { allowCross }) {
+  ensureValueNotConflict(i, val, { allowCross, pushable: thershold }) {
     const state = this.state || {};
-    const { handle, bounds } = state;
+    const { bounds } = state;
+    const handle = (i === undefined) ? state.handle : i;
```
paranoidjk: In which case `i` could be `undefiend`?

src/Range.jsx
```diff
@@ -260,23 +260,25 @@ class Range extends React.Component {
...
+    thershold = Number(thershold);
     /* eslint-disable eqeqeq */
-    if (!allowCross && handle != null) {
-      if (handle > 0 && val <= bounds[handle - 1]) {
-        return bounds[handle - 1];
+    if (!allowCross && handle != null && bounds !== undefined) {
```
paranoidjk: bounds shouldn't be `undefined` either ?

src/Range.jsx
```diff
@@ -260,23 +260,25 @@ class Range extends React.Component {
...
+      if (handle > 0 && val <= (bounds[handle - 1] + thershold)) {
```
paranoidjk: Actually, handle is a array index, so it should always greater than 0

src/Range.jsx
```diff
@@ -260,23 +260,25 @@ class Range extends React.Component {
...
+        return bounds[handle - 1] + thershold;
       }
-      if (handle < bounds.length - 1 && val >= bounds[handle + 1]) {
-        return bounds[handle + 1];
+      if (handle < bounds.length - 1 && val >= (bounds[handle + 1] - thershold)) {
```
paranoidjk: same as below, `handle < bounds.length - 1` should always be true.

Robin-front:
@paranoidjk  
Yes, It seems have any other problem after merge master. 
Same with you, any of other handle being not moveable. look like focus the same handle again.
I will check.

src/Range.jsx
Robin-front: yes, because `ensureValueNotConflict` invoked by `trimAlignValue`,  then `trimAlignValue` was invoked many place.
sometimes, only trim current handle value ( didn't pass index, like `calcValueByPos` in `createSlider.jsx` ) , but others, maybe map all handle values to trim.

src/Range.jsx
Robin-front: when init component, `trimAlignValue` invoked in `constructor`, at the time `this.state` is `undefined`, bounds also.
anyway, it must walk over this function to return origin `val`

src/Range.jsx
Robin-front: No, handle will be `null` only one moment when init component, `constructor` init state `this.state = {
      handle: null,
      recent,
      bounds,
    };`

once click any handle,  handle will be no more `null`.

Robin-front: oops, sorry, I am mistaken about you.
Here should be judge safe distance of the handle and surrounding handles.
handle is index, `handle <= bounds.length-1 && handle >= 0`, with equal.

Robin-front:
@paranoidjk 
The problem about handle cannot move is resolve already. The reason is I add extra `mouseup` event, so `onEnd` does not trigger.

And the CI test error below seems `RangeWithTooltip` error.
my local test cases are all passed.

paranoidjk:
The onEnd bug should be fixed in https://github.com/react-component/slider/commit/18d4fa33319618795f7c9cbe5045dc0d0ed33299 alreay

Robin-front:
Yes, but this `mouseup` function was add by another PR,  I think it is redundant code now. 
Because ownerDocument has `mouseup` to trigger `onEnd`  already. `onEnd` will `removeEvent` also.

src/Range.jsx
```diff
@@ -261,23 +261,24 @@ class Range extends React.Component {
     return true;
   }
 
-  trimAlignValue(v, nextProps = {}) {
+  trimAlignValue(v, handle = this.state.handle, nextProps = {}) {
```
paranoidjk: `createSlider` will call `trimAlignValue` too https://github.com/react-component/slider/search?utf8=%E2%9C%93&q=trimAlignValue&type=

paranoidjk:
And another demo of `allowCross false && pushable true` Range, both have bug right now 😅  https://github.com/react-component/slider/commit/9bbc8b13c0cb8aaf1a8a359288c2f7ad7f0a1a52

src/Range.jsx
Robin-front: Yes, but args `handle` & `nextProps` both optional. So,  I think it dosen't matter.

Robin-front:
@paranoidjk 
After merge master into my branch, it works well now. I can't reproduce.

Robin-front:
@paranoidjk 
Please checkout my lastest codes and review again.

trace: #226  #379

paranoidjk:
@Robin-front Have you test the demo you added in this PR,  pushable is still broken.

Robin-front:
This is official online example, it's broken.
offical

This is my pr, local example, it works well:
notallowcross

src/common/createSlider.jsx
```diff
@@ -171,11 +171,6 @@ export default function createSlider(Component) {
       /* eslint-enable no-unused-expressions */
     }
 
-    onMouseUp = () => {
-      this.onEnd();
-      this.removeDocumentEvents();
-    }
```
yesmeck: Why is this callback removed? Will cause a memory leak if we don't remove listeners from the document.

yesmeck: Nevermind, listeners will be removed in `onMouseDown`.

yesmeck:
@Robin-front  Examples look good, could you rebase master, then I can merge it.

src/common/createSlider.jsx
Robin-front: No, not `onMouseDown`. 
Actually, listeners will be removed in `this.onEnd`,  when another `mouseup` event  that bind on Document was trigger.

So, `mouseUp` event bind on `Slide` is repeated to another `mouseUp` bind on document.(trace: #361 )

https://github.com/react-component/slider/blob/c7f6f228d4d8a681a0d5d236d243898dda8e6421/src/common/createSlider.jsx#L161
https://github.com/react-component/slider/blob/c7f6f228d4d8a681a0d5d236d243898dda8e6421/src/Range.jsx#L111-L115

Issue #199, opened 2017-01-04
Pushable feature inaccuracy


tasiek:
There's an inaccuracy issue with pushable feature, when dragged through the "edge" of range.
After reaching the end, the final value is calculated incorrectly in some cases - it stops the handler too early.

![rc-slider-range-pushable-issue](https://cloud.githubusercontent.com/assets/4588259/21640718/5af4f6b4-d277-11e6-9184-67017cf97f39.gif)

In this case, pushable's range is set to 250.

benjycui:
It will be better if you can provide steps which can re-produce it every time.

Robin-front:
I think the reason is `mousemove` event is not continued.
![move](https://user-images.githubusercontent.com/6723674/29166576-8c1a26aa-7df9-11e7-9216-92748ef3e3ee.gif)

sometimes, the handle value will sudden skip to 1000 from 740.
But when pushing is impossible. the author reset the handle value to originalValue. That's why pushable feature inaccuracy. The logic [look here](https://github.com/react-component/slider/blob/master/src/Range.jsx#L222)

So, I think reset the handle value to limit is better, rather than originalValue.
`bounds[handle] = bounds[nextHandle] - (direction * threshold);`

paranoidjk:
@Robin-front Sounds reasonable，could you give a reproducible demo? or add some test case in your PR? Thanks.

Robin-front:
@paranoidjk okay, I will try. :)

yesmeck:
Fixed in 8.6.1

PR #305, opened 2017-08-10
pushable inaccuracy

Robin-front:
trace: #199

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -125,8 +125,7 @@ class Range extends React.Component {
     nextBounds[state.handle] = value;
     let nextHandle = state.handle;
     if (props.pushable !== false) {
-      const originalValue = state.bounds[nextHandle];
-      this.pushSurroundingHandles(nextBounds, nextHandle, originalValue);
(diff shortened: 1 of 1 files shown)


paranoidjk:
@Robin-front Thanks for your PR. I noticed that you have helped us fixed many problems. I'd like to invite you to be  a Collaborators of rc-slider, you can click this link if you are willing too https://github.com/react-component/slider/invitations

src/Range.jsx
```diff
@@ -219,7 +218,7 @@ class Range extends React.Component {
     const diffToNext = direction * (bounds[nextHandle] - value);
     if (!this.pushHandle(bounds, nextHandle, direction, threshold - diffToNext)) {
       // revert to original value if pushing is impossible
-      bounds[handle] = originalValue;
+      bounds[handle] = bounds[nextHandle] - (direction * threshold);
```
paranoidjk: `pushable` may be a boolean

Robin-front: yes, there is implicit-conversion when calculate. But of course, it's not serious. I will change to explicit.

Issue #244, opened 2017-03-23
Slider gets stuck to mouse


badaljain:
Hi, Sometimes the slider gets stuck to mouse and the slider just follows the movement of the cursor.

Its hard to reproduce the issue and only way to resolve is to reload the page.

benjycui:
Please provide a re-producible demo: http://codepen.io/benjycui/pen/aJooRE?editors=0011

And steps to re-produce it..

Forestdev:
Can confirm it's happening to me as well. Hard to reproduce. It seems it happens at random.

eddyerburgh:
@benjycui Steps to reproduce on codepen:

1. Click on handle
2. While holding click, move cursor over codepen JS window
3. Keep holding click, move cursor to opposite side of page
4. Cursor will be stuck.
5. Click on handle, if it becomes unstuck - repeat step 1 - 4

benjycui:
@eddyerburgh thx, but I can re-produce it after trying 3 times.

eddyerburgh:
![reproduced bug](http://i.imgur.com/sY4xRip.gifv "Reproduced bug")

I'm on OSX Sierra 10.12.3, chrome 57.0.2987.

Reproducible every time.

rmammina:
I get the same thing

dwighthouse:
Looks like the issue is related to the page losing the mouseup event to cancel the dragging feature on the page. Might be fixed by setting up the mousemove listener to check for the left mouse button being down every event. If any mousemove event during a dragging does not have the left mouse button active, cancel the drag immediately.

benjycui:
Sorry.. I don't understand even I look at your GIF for a long time.

If someone can re-produce this with a unit test, it will be better.

So, PR is welcomed, and please add a unit test for this. Thanks.

ContiGabriel:
Anyone found a solution for that issue?

Abcmsaj:
Video of issue: https://streamable.com/1u6k5

When dragging cursor and lifting mouse button up when NOT on the browser window (i.e. the Start Menu bar or on the desktop), the slider doesn't register that you've moused-up and let go of it, so it will continue to follow your cursor around.

Even when clicking on the slider and dragging, it doesn't reset it - only fix is refresh

paranoidjk:
@Abcmsaj I watched your demo, seems there do have a problem. But none of out team have windows computer, and i test this demo on osx, it works fine.

So, would you like to give us a PR to fix it?

[...]

PR #306, opened 2017-08-11
removeEvent on Mouseup and before addEvent

Robin-front:
trace:  #244

### DIFF
--- a/src/common/createSlider.jsx
+++ b/src/common/createSlider.jsx
@@ -116,7 +116,6 @@ export default function createSlider(Component) {
         this.dragOffset = position - handlePosition;
         position = handlePosition;
       }
-      this.removeDocumentEvents();
       this.onStart(position);
       this.addDocumentTouchEvents();
       utils.pauseEvent(e);
--- a/src/common/createSlider.jsx
+++ b/src/common/createSlider.jsx
@@ -98,6 +98,7 @@ export default function createSlider(Component) {
         this.dragOffset = position - handlePosition;
         position = handlePosition;
       }
+      this.removeDocumentEvents();
       this.onStart(position);
       this.addDocumentMouseEvents();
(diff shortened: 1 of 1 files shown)


monsieurnebo:
Any news about this PR?

paranoidjk:
Sorry for the late reply. Thanks for you pr, I will take a look right now.

src/common/createSlider.jsx
```diff
@@ -98,6 +98,7 @@ export default function createSlider(Component) {
         this.dragOffset = position - handlePosition;
         position = handlePosition;
       }
+      this.removeDocumentEvents();
```
paranoidjk: This will remove both TouchEvent and MouseEvent listener.

In mobile device, if there is no touchmove, the event will trigger follow `touchstart --> touchend --> onmousedown --> onmouseup`

Robin-front: Eh..sorry. I am confused. Maybe I miss something. 
clear event before addEvent, there is `touchmove` again.  I don't know when that will happen.

Robin-front: You mean I remove both TouchEvent and MouseEvent listener, so there is no more touchmove, the event will trigger follow `touchstart --> touchend --> onmousedown --> onmouseup`; 
But I just remove last event listener. In 103 lines, it add new event listener for this action. There is touchmove.

This bug happened when your mouse or touch move to another document(e.g. parent document with iframe).  The event had bind in origin document, can not be remove correctly. So we have to check with extra func, to ensure we have a clean context.

Or I misunderstood.  :D

mpeula-pe:
Any progress in this PR?

paranoidjk:
close https://github.com/react-component/slider/issues/244

jitao13:
The problem will be repeated on the PC.

PR #307, opened 2017-08-12
calcOffset should always return a number

sunny-g:


paranoidjk:
Have you came across any problem or find a bug?

Please give more detail info, thanks.

src/common/createSlider.jsx
```diff
@@ -215,7 +215,7 @@ export default function createSlider(Component) {
     calcOffset(value) {
       const { min, max } = this.props;
       const ratio = (value - min) / (max - min);
-      return ratio * 100;
+      return isNaN(ratio) ? 0 : ratio * 100;
```
paranoidjk: I don't konw in what case ratio could be `NaN`

- `max - min` should always more than `0` if uses give legal props

- `value - min` should always no less than `0`

So i think ratio should be always no less than `0` and no more than `1`

benjycui:
Close for no response.

sunny-g:
Sorry, can't remember the exact case in which this occurred, but it did happen and this PR fixed it. I believe the bug was that when the current `value` was out of bounds of the previously set `bounds`, `value` became `null`.

In either case, wouldn't you want some default implementation like this to avoid type conversion errors?

benjycui:
No useless code, it will be confusing.

If you can re-produce, please provide an online demo, thanks.

PR #312, opened 2017-08-21
Add support for 'alwaysVisible' property

uberstudio:
This adds support for when we want the tooltip to be always visible

### DIFF
--- a/src/createSliderWithTooltip.jsx
+++ b/src/createSliderWithTooltip.jsx
@@ -49,7 +49,7 @@ export default function createSliderWithTooltip(Component) {
           prefixCls={prefixCls}
           overlay={overlay}
           placement={placement}
-          visible={!disabled && (this.state.visibles[index] || dragging)}
+          visible={!disabled && tipProps.alwaysVisible || (this.state.visibles[index] || dragging)}
           key={index}
         >
           <Handle


glorp47:
+1 for having this type of functionality ...

paranoidjk:
Thanks for your PR. But you can do this by custom handle, eg: https://github.com/react-component/slider/blob/master/examples/handle.js#L34

hellofantastic:
The docs or readme says Range component supports handle but id I have <Range  handle={(props=>{this.tooltipHandle(props) }}  nothing fires... but if I add it to Slider it does work... insight? 

```javascript
...
tooltipHandle = (props) => {
        const { value, dragging, index, ...restProps } = props;
        console.log('tooltip');
        return (
          <Tooltip
            prefixCls="rc-slider-tooltip"
            overlay={value}
            visible={true}
            defaultVisible={true}
            placement="bottom"
            key={index}
          >
            <Handle value={value} {...restProps} />
          </Tooltip>
        );
    };
...
<Range
                                className="mt-3"
                                style={this.props.rangeStyle}
                                min={0}
                                max={100}
                                value={this.state.rangeValues[rangeConfig.field]}
                                
                                handle={(props)=>{ console.log(props); }}
                                onChange={(value) => { this.onRangeChange(rangeConfig.field, value); }}
                                onAfterChange={(value) => {this.onAfterRangeChange(rangeConfig.field, value);}}
                            />
                            <Slider  min={0}
                                max={100}
                                handle={(props)=>{ console.log(props); }}/>
```

uberstudio:
Just do
```handle={tooltipHandle}```

Looks like you're just logging there instead of using the handle function you wrote :)

Issue #319, opened 2017-09-13
Overlapping marks on max point cannot be moved once placed


jeznag:
![issue with overlapping marks](https://user-images.githubusercontent.com/1034890/30405103-5e574406-992d-11e7-8db5-92f4cef84c46.gif)

PR #320, opened 2017-09-13
fix($range): Fix issue with overlapping points

jeznag:
Bug when trying to move overlapping points

#319


paranoidjk:
Nice catch. Thanks 👍

Issue #323, opened 2017-09-26
PropTypes and createClass warnings


primozs:
It looks like Popup and Trigger packages dependencies should be upgraded.
with rc-slider version 8.3.1 there are still react warnings for createClass and PropType usage.

primozs:
This is not resolved.
When I install "rc-slider": "^8.3.2",
at least ├─ rc-trigger@1.11.5 is old one and I still get PropTypes warnings. Still unable to upgrade 

```
success Saved 11 new dependencies.
├─ add-dom-event-listener@1.0.2
├─ component-classes@1.2.6
├─ component-indexof@0.0.3
├─ css-animation@1.4.1
├─ dom-align@1.6.5
├─ rc-align@2.3.4
├─ rc-animate@2.4.1
├─ rc-slider@8.3.2
├─ rc-tooltip@3.5.0
├─ rc-trigger@1.11.5
└─ rc-util@4.1.1
```
```
yarn why rc-trigger
yarn why v1.1.0
[1/4] 🤔  Why do we have the module "rc-trigger"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
info This module exists because "rc-slider#rc-tooltip" depends on it.
info Disk size without dependencies: "12.09MB"
info Disk size with unique dependencies: "21.2MB"
info Disk size with transitive dependencies: "36.19MB"
info Number of shared dependencies: 22
```

benjycui:
Make sure that you haven't lock deps.

primozs:
I have uninstalled and installed again then I reported the problem. Also I have checked the repos chain:

Rc-slider packgage.json has dependency to rc-tooltip 3.4.3 current published version is 3.5.0.
Unfortunately even rc-tooltip master has dependency to "rc-trigger": "1.x". Current version of rc-trigger where I presume this is resolved is 2.1.0.

Because this is a family of modules I presume its ok to report the problem at the top? I have checked and also rc-tooltip has already an issues about react v16. 

The thing is, even when that will be resolved it will still not be here in rc-slider.

Thank you for help and great module!!

PR #325, opened 2017-09-27
Upgrade rc-tooltip version

aray12:
The rc-tooltip -> rc-trigger -> rc-align dependency is causing React 16 upgrade warnings. As far as I can tell, all of the libraries have been updated and we just need to upgrade rc-tooltip.

There might be some issues with rc-trigger and its portal implementation

#323


benjycui:
Just `rm -rf node_modules && npm i`, make sure that you never lock deps.

PR #326, opened 2017-10-03
change event binding to slider ownerDocument

polkpolio1:
I noticed a problem that events binding to document. It has problem in frames that has they own document. It is better to use _el.ownerDocument_ to add all slider event listeners.


benjycui:
Agree, but could you just show the problem with online demo? Thanks.

polkpolio1:
here it is

polkpolio1:
Here it is:
https://codepen.io/anon/pen/gGQyvP?editors=1011
demo

benjycui:
8.3.2

PR #327, opened 2017-10-06
SPIKE: Pushable/combine with 0 thresholds

polizz:
This is basically a feature request with some POC code to get your thoughts.

We use your control in the Range mode with multiple initial default values. We like pushable, but prefer not to have the threshold between handles. In this way, we want to allow the handles to collide, so that the handles you run into, push when they are run into, not before. Thus, you would go from two (or more) values to a single value when you collide with the other handle.

From what I can tell from your implementation, the handles are simply visualizations of an underlying array of (presumed) unique values, so this test currently behaves as though it combines the handles and you **cannot** select the other handle again. Without another widget to split them or reset the control state, I'm guessing there is no way to go back to your previous multiple handles.


yoyo837:
Rebase please.

PR #329, opened 2017-10-07
Only update values that are out of range.

lukashambsch:
Range component should only update the values that are out of range.

Issue: https://github.com/react-component/slider/issues/328

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -64,13 +64,7 @@ class Range extends React.Component {
 
     if (bounds.some(v => utils.isValueOutOfRange(v, nextProps))) {
       const newValues = value.map((v) => {
-        if (v < nextBounds[0]) {
-          v = nextBounds[0];
-        }
-        if (v > nextBounds[1]) {
-          v = nextBounds[1];
-        }
-        return v;
+        return utils.ensureValueInRange(v, nextProps);
       });
       this.props.onChange(newValues);
     }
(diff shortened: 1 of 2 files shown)


src/Range.jsx
```diff
@@ -61,8 +61,18 @@ class Range extends React.Component {
     if (nextBounds.length === bounds.length && nextBounds.every((v, i) => v === bounds[i])) return;
 
     this.setState({ bounds: nextBounds });
+
     if (bounds.some(v => utils.isValueOutOfRange(v, nextProps))) {
-      this.props.onChange(nextBounds);
+      const newValues = value.map((v) => {
```
Robin-front: maybe this is what [here(`ensureValueInRange`)](https://github.com/react-component/slider/blob/0a07adac3694f5881ecf9dffcd4831ba6fafeff9/src/utils.js#L53) do

tests/Range.test.js
```diff
@@ -79,6 +79,14 @@ describe('Range', () => {
     expect(wrapper.find('.rc-slider-handle').length).toBe(2);
   });
 
+  it('should only update bounds that are out of range', () => {
+    const props = { min: 0, max: 10000, value: [0.01, 10000], onChange: jest.fn() };
+    const range = mount(<Range {...props} />);
```
Robin-front: I think here should add step for calc precision, `const range = mount(<Range {...props} step={0.01} />);` , if no step, the precision will be `1`.  Maybe that is why this test case fail before pr.

src/Range.jsx
```diff
@@ -61,8 +61,12 @@ class Range extends React.Component {
     if (nextBounds.length === bounds.length && nextBounds.every((v, i) => v === bounds[i])) return;
 
     this.setState({ bounds: nextBounds });
```
yesmeck: Should `nextBounds` have the same value as following `newValues`?

lukashambsch: If the value is already within the range it shouldn't be changed.

Issue #333, opened 2017-10-11
Upgrade to react 16 already on master, can a tag please be cut and published?


caguthrie:
Would love to use this with React 16 in my project.  Thanks

benjycui:
Just upgrade react in your project, we don't need to publish new version of rc-slider.

caguthrie:
With react@16, dependency rc-tooltip@3.4.2 is incompatible as it uses `React.PropTypes`.  While `"rc-tooltip": "^3.4.2"` is listed in package.json which should resolve to a good version normally with the `^`, if a user is using a yarn.lock, package.lock, or similar lockfile with rc-tooltip@3.4.2, nothing will appear wrong when dependencies are resolved.  If you want to reopen this ticket I'm happy to raise a PR.

benjycui:
> If you want to reopen this ticket I'm happy to raise a PR to change this to "rc-tooltip": "^3.4.3", the first version to no longer use React.PropTypes.

OK, just PR.

PR #336, opened 2017-10-18
Increasing minimum version of rc-tooltip dependency that is compliant with react 16

caguthrie:
Addresses https://github.com/react-component/slider/issues/333#issuecomment-337313098


benjycui:
rc-slider@8.3.3

PR #340, opened 2017-10-20
Do not render empty mark labels

uwolfer:
Use-case: I have a slider bar for percent values. I use marks for
every allowed value (e.g. values from 2 to 5 are not allowed).
Displaying every mark label would flood the UI. So there are mark labels
displayed only for every 10th step, which also works fine without this
change (using: {10: '10%', 11: ''}).
This change just optimizes that there are no empty DOM nodes rendered.

### DIFF
--- a/src/common/Marks.jsx
+++ b/src/common/Marks.jsx
@@ -17,6 +17,14 @@ const Marks = ({
 
   const range = max - min;
   const elements = marksKeys.map(parseFloat).sort((a, b) => a - b).map(point => {
+    const markPoint = marks[point];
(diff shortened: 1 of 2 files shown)


paranoidjk:
Sorry, could you explain what problems does render empty mark labels will cause ?

IMO, empty string actually  is a valid React Element, if we add too many such check in a React Component, it will be redundancy.

uwolfer:
It's not a bugfix. It's just an optimization. This change just prevents rendering useless empty dom nodes.

PR #364, opened 2017-11-21
fix handleStyle with createSliderWithTooltip

Orange-C:
porps.handleStyle could be array or object now. 
**fix bug**: handleStyle[0] work with all handles

by the way, how about supporting porps.handle when using createSliderWithTooltip?


yesmeck:
Could you add a test case, then I can merge it.

Orange-C:
@yesmeck It's too late to merge this commit : ( , looks like someone in your team fixed the problem during these months , just close this. 
Hope you check pr more frequently in the future : )

PR #369, opened 2017-11-24
Remove visibility property from rc-slider-track

7brend7:
Had an issue when range component placed inside dropdown which controlled by visibility property. And in result when dropdown is hidden rc-slider-track still visible.

### DIFF
--- a/src/common/Track.jsx
+++ b/src/common/Track.jsx
@@ -13,11 +13,10 @@ const Track = (props) => {
   };
 
   const elStyle = {
-    visibility: included ? 'visible' : 'hidden',
     ...style,
     ...positonStyle,
   };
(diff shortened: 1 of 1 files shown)


yesmeck:
Could you demonstrate your issue on https://codesandbox.io/?

7brend7:
emm... so fast response (((

here is an example for issue: https://codesandbox.io/s/v0pj0r5zj5

Actually - you are able to click on _Track_ even with opacity:0
I really don't know why _Track_ component uses 'visibility' style, but it can break some styles, especially when you use slider in some popup element.

also... https://www.w3.org/TR/CSS22/visufx.html#visibility

yesmeck:
Make sense, thanks!

Issue #213, opened 2017-02-02
Marks: usability issue


websash:
Nice component but there's a usability issue when using marks. The area with labels under the slider should be used for setting values **defined by those labels only** (not for arbitrary values). 

In the example below
![screen shot 2017-02-02 at 12 35 39 pm](https://cloud.githubusercontent.com/assets/679341/22547392/10d90266-e94a-11e6-863a-bb9aa7f24bdf.png)
when I click 25 I want 25 (not 24 or 26 or anything close). The way it works now kind of defeats the purpose of having these predefined marks in the first place. You have to be very precise with your mouse to select them which is pointless

benjycui:
Yep, it's a little weirdo. But I cannot find a better solution. Because I cannot tell what users are going to do, drag or just click on mark? 

As you know, browser will trigger `mousedown` and `click` when use click on Slider.

benjycui:
Close for no response.

williamswgb:
About this issue, isn't it better if clicking the marks will change the value and move the handle to the corresponding mark value? The point of having marks on the slider is to let the user change the slider value easier by clicking on markers instead of dragging the slider all the way in order to get the right value.

PR #376, opened 2017-12-07
Add click handler to mark in order to change the slider value to mark value when the mark label is clicked

williamswgb:
This is to fix the usability issue of clicking mark label sometimes unable to update the slider input with the correct mark value.
https://github.com/react-component/slider/issues/213.


yesmeck:
LGTM, could you add a test case?

PR #379, opened 2017-12-15
Fixed issue where controlled Range sliders with allowCross=false and …

Oddsor:
…more than two handles would change values of handles that aren't currently being dragged

This fully(?) solves issue https://github.com/react-component/slider/issues/226

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -268,13 +268,13 @@ class Range extends React.Component {
     return utils.ensureValuePrecision(valNotConflict, mergedProps);
   }
 
-  ensureValueNotConflict(val, valsHandle, { allowCross }) {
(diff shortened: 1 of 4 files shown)


Oddsor:
Added a test case that describes the original issue I encountered with issue #226, along with a proposed fix.

This is basically my first open source contribution, so any feedback would be appreciated!

Robin-front:
similar to #304 , cause wrong handle, didn't pass `index` when `map`

Oddsor:
Very similar indeed! It looks like your pull request is missing a test case for this particular issue however. Maybe I could limit the pull request to this.

yesmeck:
#304 has been merged, try 8.6.1

Issue #353, opened 2017-10-31
Can't set tabIndex on handle


gvincentlh:
We need to be able to disable tabbing into the `rc-slider-handle` by setting `tabIndex="-1"`. Unfortunately it looks like tabIndex on the handle is currently hard-coded to "0", with no ability to override this value. https://github.com/react-component/slider/blob/76d4cf3567c48a43922f6a8c00f5b892e62d41e3/src/Handle.jsx#L28

Could you please add the ability to override `tabIndex` on the handle?

paranoidjk:
It's for keyboard accessibility, could you explain why you need to disable it ? https://github.com/react-component/slider/pull/282

gvincentlh:
@paranoidjk does the **why** really matter? We'd just like to have control over the `tabIndex` value on the handle. We want to create an explicit tab order on multiple page elements. Thanks.

paranoidjk:
Then we maybe could add a api `tabIndex: boolean | number (default to be 1, false to disable)`

Would you like PR?

gvincentlh:
@paranoidjk sounds good, yes will submit a PR tomorrow.

gvincentlh:
@paranoidjk see the MDN docs on tabIndex here https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

We should probably default it to "0" as it is now, but allow it to be overridden via props. I'll expose a `tabIndex` prop on the Handle component.

PR #381, opened 2017-12-19
Allow tabIndex to be set explicitly on Handle

gvincentlh:
Basic implementation of `tabIndex` prop on the Handle component. Both Slider.js and Range.js also implement `tabIndex,` but in both cases it's simply passed through to the Handle. I've also updated the test suite with some simple checks to make sure `tabIndex` gets set properly on Handle from its parent. I've updated the README doc as well.

The default value for `tabIndex` is preserved (it's still "0"), so this change should not break any existing uses.

This is a fix for https://github.com/react-component/slider/issues/353

cc: @paranoidjk

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ The following APIs are shared by Slider and Range.
 | ------------ | ------- | ------- | ----------- |
 | defaultValue | number | `0` | Set initial value of slider. |
 | value | number | - | Set current value of slider. |
+| tabIndex | number | `0` | Set the tabIndex of the slider handle. |
 
 ### Range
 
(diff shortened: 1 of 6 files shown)


gvincentlh:
Thanks @paranoidjk and @yesmeck ! When can we expect a new release?

gvincentlh:
@paranoidjk @yesmeck could we get another release with that change in it please?

yesmeck:
Released rc-slider@8.6.0

gvincentlh:
Thanks!

gvincentlh:
@yesmeck I'm not seeing my changes reflected in the `8.6.0` **dist** files. Were they regenerated?

yesmeck:
Yes, search `tabIndex || 0` in this file https://unpkg.com/rc-slider@8.6.0/dist/rc-slider.js

Shardendu-kantha:
I updated to 8.6.0 version, but passing tabIndex={-1} is still not setting tabindex=-1 for handles for Range.

`<Range value={[minPrice, maxPrice]}
                    min={this.minPriceRange}
                    max={this.maxPriceRange}
                    onChange={this._onPriceChange}
                    onAfterChange={this._onPriceAfterChange}
                    step={null}
                    marks={this.state.priceSliderMarksObj}
                    tabIndex={-1}
                    />`

Robin-front:
@Shardendu-kantha 
`<Range>` expect an array for `tabIndex` prop. Different from `<Slider>`

https://github.com/react-component/slider/blob/7a021d451521615083790d81a88d21a6d60879ff/src/Range.jsx#L23

Shardendu-kantha:
@Robin-front thank you so much. Got the solution. cheers

Issue #377, opened 2017-12-07
Range Doesn't Support Custom Tooltip


lgants:
I copied the example in the docs and added a Range component that uses the same tooltip as the Slider. The custom tooltip on Slider works as anticipated. However, the tooltip on both Range components is identical despite one of the Range components receiving a handle prop with custom tooltip. Thoughts?

```
import 'rc-slider/assets/index.css';
import 'rc-tooltip/assets/bootstrap.css';
import React from 'react';
import Tooltip from 'rc-tooltip';
import Slider from 'rc-slider';

const createSliderWithTooltip = Slider.createSliderWithTooltip;
const Range = createSliderWithTooltip(Slider.Range);
const Handle = Slider.Handle;

const handle = (props) => {
  const { value, dragging, index, ...restProps } = props;
  return (
    <Tooltip
      prefixCls="rc-slider-tooltip"
      overlay={value}
      visible={true}
      placement="top"
      key={index}
    >
      <Handle value={value} {...restProps} />
    </Tooltip>
  );
};

const wrapperStyle = { width: 400, margin: 50 };

export default () => {
    return (
        <div>
            <div style={wrapperStyle}>
                <p>Slider with custom handle</p>
                <Slider min={0} max={20} defaultValue={3} handle={handle} />
            </div>
            <div style={wrapperStyle}>
                <p>Range with custom handle</p>
                <Range min={0} max={20} defaultValue={[3, 10]} handle={handle} />
            </div>
            <div style={wrapperStyle}>
                <p>Range with custom handle</p>
                <Range min={0} max={20} defaultValue={[3, 10]} tipFormatter={value => `${value}%`} />
            </div>
      </div>
    )
}
```

aranair:
I have a custom tooltip that is working for me. You can try importing Range instead?

```
import Slider, { Range } from 'rc-slider';
```

hurrtz:
Have you tried logging anything in your handle function to ensure that the function gets used? In my case the prop "handle" does not get triggered if the range is used as a createSliderWithTooltip(range). Guess, this is a bug... Theoretically you could work around this via the prop "handleStyle" and supply an array with objects containing css definitions, one object per handle, but there, also, is a bug where only the first object is being used and applied over all handles. Sorry for all the bad news, though.

aranair:
Hmm, I’m 100% certain it’s working for me because of a bunch of custom date formatting in it.

bcvazquez:
[...]

PR #383, opened 2017-12-26
Added forceVisible parameter to tipProps.

oziian:
If need always to show toolTip value.

I want to show toolTip always. But I can't change this state. It works only on hover. 
https://github.com/react-component/slider/blob/master/src/createSliderWithTooltip.jsx#L61
I added this changes because custom toolTip doesn't work.
https://github.com/react-component/slider/issues/377

### DIFF
--- a/src/createSliderWithTooltip.jsx
+++ b/src/createSliderWithTooltip.jsx
@@ -40,6 +40,7 @@ export default function createSliderWithTooltip(Component) {
         prefixCls = 'rc-slider-tooltip',
         overlay = tipFormatter(value),
         placement = 'top',
+	    forceVisible = forceVisible || false,
         ...restTooltipProps,
       } = tipProps;
 
@@ -49,9 +50,10 @@ export default function createSliderWithTooltip(Component) {
           prefixCls={prefixCls}
           overlay={overlay}
           placement={placement}
-          visible={!disabled && (this.state.visibles[index] || dragging)}
+          visible={(!disabled && (this.state.visibles[index] || dragging)) || forceVisible}
           key={index}
         >
+
           <Handle
             {...restProps}
             style={{
--- a/src/createSliderWithTooltip.jsx
+++ b/src/createSliderWithTooltip.jsx
@@ -40,7 +40,7 @@ export default function createSliderWithTooltip(Component) {
         prefixCls = 'rc-slider-tooltip',
         overlay = tipFormatter(value),
         placement = 'top',
-	    forceVisible = forceVisible || false,
+        forceVisible = forceVisible || false,
         ...restTooltipProps,
       } = tipProps;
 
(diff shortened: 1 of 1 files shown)


tmikoss:
Would love this, or really any other way of forcing tooltips visible at all times.

yesmeck:
I think we can just call it `visible`?

oziian:
Renamed to `visible`.

yesmeck:
Thanks!

mradzwilla:
Can someone provide an example of how to implement this feature?

oziian:
@mradzwilla  You need just add `visible={true}`. 
But I didn't find my code in the npm repository. The repository published a month ago, and my code merged at the same time.
I think we will need to wait until @yesmeck publish code into npm again.

```js
import React from 'react';
import PropTypes from 'prop-types';
import Slider from 'rc-slider';

const createSliderWithTooltip = Slider.createSliderWithTooltip;
const Range = createSliderWithTooltip(Slider.Range);

export class RangePrice extends React.Component {
	constructor (props) {
		super(props);
		this.state = {
			min: props.min,
			max: props.max
		};
		this.onChange = this.onChange.bind(this);
	}

	onChange (value) {
		this.props.onChange({min: value[0], max: value[1]})
	};

	componentWillReceiveProps(nextProps) {
		if (nextProps.max !== this.props.max) {
			this.setState({max: nextProps.max});
		}
		if (nextProps.min !== this.props.min) {
			this.setState({min: nextProps.min});
		}
	}

	render() {
		let defaultValue = [this.state.min, this.state.max];
		return(
			<div className="range-price">
				<Range min={this.state.min}
				       max={this.state.max}
				       defaultValue={defaultValue}
				       onChange={this.onChange}
				       visible={true}
				       tipFormatter={value => value} />
			</div>
		)
	}
}

RangePrice.propTypes = {
	min: PropTypes.number,
	max: PropTypes.number,
	onChange: PropTypes.func,
};

RangePrice.defaultProps = {
	min: 0,
	max: 100,
	onChange: () => {},
}; 
```

lunatik-210:
I suppose it is in tipsProps like that:

`tipProps={{visible: true}}`

Otherwise, it's not showing up..

But there's a consequence of that, tooltips are not updating with a scroll, and will show randomly on the page until you will mouse focus on the slider itself. Are there any ways to overcome it? Or should i open a bug?

yesmeck:
@lunatik-210 Please file a new issue and try to reproduce it on https://codesandebox.io?

computerjazz:
I'm seeing the buggy behavior in my app as well. 
rc-slider-bu


However, when I created an example it works as expected:
https://codesandbox.io/s/j2x418y2yw

okrookrps:
> I'm seeing the buggy behavior in my app as well.
> rc-slider-bu
> 
> However, when I created an example it works as expected:
> https://codesandbox.io/s/j2x418y2yw

Here is the reproducible example
https://codesandbox.io/s/yv39p3qmx

PR #398, opened 2018-02-20
Remove pauseEvent

### LINKED ISSUE #349: Can I simulate the "onFocus" event?
Hello. In my project we save the data in the server when the user leaves the form element (input type). But, if I am focused on an input text, and I click on the Slider, in the input text doesn't execute the "blur" event (then I can't save the data). Please see the attached file.

oct-26-2017-12-48-pm

LandonSchropp:
Pausing events can break the normal browser behavior for other inputs in the DOM. For example, focused inputs will fail to blur when interacting with a slider. This can cause unexpected user behavior in the browser.

I'm opening this PR to facilitate discussion, but if you'd like to merge that's totally okay.

Fixes #349

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -115,7 +115,6 @@ class Range extends React.Component {
   }
 
   onMove(e, position) {
-    utils.pauseEvent(e);
     const props = this.props;
     const state = this.state;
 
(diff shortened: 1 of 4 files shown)


yesmeck:
pauseEvent can not be removed simply. For example in `onKeyboard` callback, we do need prevent default behavior when user press <kbd>Page Down</kbd>, I'll look into this issue. Thanks anyway.

LandonSchropp:
Thanks for taking a look at this issue. The use case mentioned in #349 is
definitely a problem for me as well.

PR #401, opened 2018-02-25
Add click to focus

### LINKED ISSUE #335: Cannot control slider via Keyboard
Hi, I cannot control the slider via keyboard.

I checked the example here with Chrome v61 by pressing on the handle and then trying to move it with the arrow keys:
http://react-component.github.io/slider/examples/handle.html

I think no keyboard events are fired at all, because the outer `div` can never have focus (no `tab-index`) and prevents the handle component from getting focus because it calls [`utils.pauseEvent(event);` on `Mousedown`](https://github.com/react-component/slider/blob/master/src/common/createSlider.jsx#L107).

yesmeck:
Fix #335 #349

### DIFF
--- a/assets/index.less
+++ b/assets/index.less
@@ -74,6 +74,11 @@
       box-shadow: 0 0 0 5px tint(@primary-color, 50%);
       outline: none;
     }
+
+    &-click-focused:focus {
+      border-color: tint(@primary-color, 50%);
+      box-shadow: unset;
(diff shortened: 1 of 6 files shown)


apazzolini:
@yesmeck Any idea when this will get released? It's been merged in for a little while.

Zodiase:
Love to see this released as well.

PR #425, opened 2018-05-03
Set dotStyle and activeDotStyle per mark

### LINKED ISSUE #206: Feature request: Allow marks to receive custom styling
Would be useful to be able to style or otherwise pass a CSS function to individual marks. Looks like it could be easily implemented in the styles array on the mark component.

ibash:
Fixes #206 

This allows setting custom dot styling on a per mark basis. Tried to be as least-intrusive as possible, but let me know if you'd prefer a different approach.


ibash:
cc @yesmeck what do you think?

yoyo837:
Rebase please.

regeter:
So sad that this simple PR never made it.
I've been fighting for an hour trying to style a single dot.

Issue #308, opened 2017-08-15
Slider reacting to wrong gesture on touch devices


nealoke:
First of all thank you for this over the moon great component 🔥 

My issue is relevant on touch screens and I'm not sure if it is a problem of the component itself but I would love some suggestions / fixes for this issue. 

My slider is horizontal and I have a few rendered below each other. When trying to navigate up and down the page and accedently touching a slider it also moves and thus blocks my navigation. 

The ideal solution I can think of is a `prop` that could turn off this behaviour like `interactHorizontal` and `interactVertical`. 

**Example of the issue**
![navigation issue](https://user-images.githubusercontent.com/17417414/29314735-a964d10a-81bf-11e7-97b1-22cac2ce5a1f.gif)

Robin-front:
It's a feature. If you touch/click anywhere of axis, slider will be change value and addEventListener to observe touchmove gesture. 

Of course, obviously you don't need it at the time. I think it should be add a prop to configure whether the feature is effective.

paranoidjk:
Yep, right now we do not lock gesture direction.

PR is welcome.

nealoke:
@paranoidjk is this possible at all? :) Just wondering how I could help, but honestly I would have no idea on how to to such a thing 😨

paranoidjk:
https://github.com/react-component/slider/blob/master/src/common/createSlider.jsx#L169

@nealoke  calculate gesture move direction in touchmove, eg: https://github.com/react-component/gesture/blob/master/src/index.tsx#L341

nealoke:
So it is already available in this component? or am I missing something? :)

DieterHolvoet:
Any progress on this issue?

DieterHolvoet:
Chances are I'll get some time at work to fix this issue during the coming week, I'll keep you posted.

Idicious:
Having the same issue with horizontal scrolling + vertical sliders.

DieterHolvoet:
I already fixed this, just need to take the time to test some more and make a pull request

gyzerok:
Hello everyone!

The issue with the slider and touch screen seem to still exist. Are there any updates on the solution? Is there a workaround for it?

PR #427, opened 2018-05-06
Touch direction locking

DieterHolvoet:
Fixes issue #308. There's still 3 tests failing and all are related to `dragOffset`, but I don't really know what's wrong, so it would be nice if someone could have a look.

#### Moved a big part of the code in onTouchStart to onFirstTouchMove to be called once within onTouchMove
**Reason:** It is impossible to determine the direction of the swipe in onTouchStart because there's only one touch point to work with. Because of that, we have to postpone the initialization of the component to the first call of onTouchMove, because at that point we have a second touch point to calculate the swipe direction with. If the swipe is in the wrong direction, the component will never be initialized, and page scroll will work as expected.

#### Added dragging state with matching class `${prefixCls}-dragging`
**Reason:** because of the `touch-action: none` and `touch-action: pan-x` styles, page scrolling is interrupted regardless of what the component does. I added a new css class to set these styles after the component initialized to fix scrolling issues.

### DIFF
--- a/assets/index.less
+++ b/assets/index.less
@@ -27,7 +27,6 @@
   padding: 5px 0;
   width: 100%;
   border-radius: @border-radius-base;
-  touch-action: none;
   .borderBox();
 
   &-rail {
@@ -58,7 +57,6 @@
     border-radius: 50%;
     border: solid 2px tint(@primary-color, 50%);
     background-color: #fff;
-    touch-action: pan-x;
 
     &:hover {
       border-color: tint(@primary-color, 20%);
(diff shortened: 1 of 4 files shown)


DieterHolvoet:
I should also say that the horizontal slider with my changes has been used on a production site for about three months, everything working perfectly: https://www.klj.be/spelen/spelen-maken-of-vinden/spelendatabank

dislick:
Your changes work very well for me, without them the slider is basically unusable on a mobile device where the user needs to scroll up and down the page. Could you please have a look at this @yesmeck ?

grahamcracker:
@zombieJ I'm interested in fixing this scroll issue as well - if I'm able to fix the unit tests, would this be a candidate for merging?

zombieJ:
hi @grahamcracker , since I don't familiar with touch devices. I may need others to help check on this PR.
But yes, PR is welcomed : )

0x5e5e:
any updates on this?

yoyo837:
Rebase please.

Issue #356, opened 2017-11-08
Set null initial value


champeng:
Can we set null value initially? I mean I don't want to show any active step on load. Right now it forces the value to assume defaultValue if it is not set or set outside bounds of [min, max].

cedricmessiant:
I would like to implement this. Did someone already worked on this?

cedricmessiant:
It's done in https://github.com/react-component/slider/pull/447

PR #447, opened 2018-07-10
Feature: add the possibility to use null as value for Slider

cedricmessiant:
As discussed in #356 I added the possibility to use `null` as value for `Slider` component (see example in /examples/slider.html ).


cedricmessiant:
@yesmeck Can you take a look at it / merge please?

PR #451, opened 2018-07-21
Allow to pass array as tipProps

### LINKED ISSUE #450: Tooltips cannot be styled separately for distinct handles
In a project I have to implement a slider to set a minimal and a maximal value for filtering. According to the design the first handle's tooltip should be positioned bottomLeft, the second bottomRight. Currently that is not possible with rc-slider.

tolads:
fixes #450


yoyo837:
Rebase please.

PR #476, opened 2018-10-02
Use ClientX instead of PageX

LiborOl:
Slider using PageX does not work correctly, if the Slider is in element with position:fixed.


yoyo837:
Rebase please.

PR #498, opened 2018-11-17
Add closest bound to onBeforeChange event

lutskanu:
Add closest bound identifier to onBeforeChange event (0 for left and 1 for right).


yoyo837:
Rebase please.

PR #506, opened 2018-12-18
Remove reset handler state onEnd

### LINKED ISSUE #503: Bug with range slider and arrow keys
Hello,

There is a bug with range slider and arrow keys. After dragging the slider if I press arrow key it would immediately crash.

bug with range slider

Seems that after dragging we lose a track on what index we are handling with. I can try to fix this when I find some time next days.

dajk:
Fixes #503 

I found this was added here: https://github.com/react-component/slider/commit/ac0e5cb86fa56e956d5cd6d670473f9c4b2c44a1 
In this case, more dealbreaker than the solution. If somebody can please review, so we can have stable thing.

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -115,9 +115,6 @@ class Range extends React.Component {
   }
 
   onEnd = () => {
-    this.setState({
-      handle: null,
-    });
     this.removeDocumentEvents();
     this.props.onAfterChange(this.getValue());
   }


yoyo837:
Rebase please.

dajk:
The file doesn't existing anymore and the PR can be closed after 5 years 🤷‍♂️

PR #510, opened 2018-12-27
Critical Bug Fix: Keyboard listeners for range slider

M-Izadmehr:
When using [the range slider ](http://react-component.github.io/slider/examples/range.html), the current handle is not determined correctly when using the keyboard to move the handles.
As a result, when using range-slider by using **arrow keys** the whole component crashes.
I made a quick fix to use the recent handle if the current handle is not found.

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -214,7 +214,7 @@ class Range extends React.Component {
     const { state, props } = this;
     const nextBounds = [...state.bounds];
     nextBounds[state.handle] = value;
-    let nextHandle = state.handle;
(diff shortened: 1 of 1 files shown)


GiselaMD:
When this fix is going to be merged?

ztplz:
image
It will still crash. 
image
`state.handle` => `state.handle || state.recent`

ztplz:
But I found a new problem. 
untitled7
I think the original logic should be fixed.

yoyo837:
Rebase please.

PR #513, opened 2019-01-08
Allow passing children to Handle component

### LINKED ISSUE #436: Customize to insert value inside Handle
Hi,
I want know if it's possibile to customize it to add the current value inside the handle, for example like this:

schermata 2018-05-22 alle 18 12 44

draperunner:
Fixes #436 

This allows the developer to pass custom components to be rendered within the Handle. I have added an example where the value of the slider is shown within the handle.


draperunner:
My bad. This already works because children are passed through `restProps` .

PR #571, opened 2019-05-14
Support coloring track < minRange by adding a zero-index track / trackStyle

justinpincar:
It appears there is currently no way to color the minimum track in a range. To solve this, I created an additional zero-index track which uses the existing `trackStyle` property.

This would change the coloring for existing implementations so I don't expect it to be merged, but figured it would be helpful to someone :-)

Before:

Screen Shot 2019-05-14 at 15 55 55

After:

Screen Shot 2019-05-14 at 15 56 14

### DIFF
--- a/src/Range.jsx
+++ b/src/Range.jsx
@@ -386,24 +386,37 @@ class Range extends React.Component {
       })
     });
 
+    const getTrackClassName = (i) => classNames({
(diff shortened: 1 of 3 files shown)


yoyo837:
Rebase please.

PR #578, opened 2019-06-06
Add a way to determine event source from onChange event

aleqsio:
We've encountered a usecase where we needed to handle onChange event differently depending if the click originated from a label or a slider element.
This fork adds a `source` to the onChange argument dictionary, that takes values `keyboard`|`label`|`slider` and allows to handle such a usecase.


haxxxton:
@aleqsio any chance you could amend your PR to only include the changes you intended to commit rather than ones that have been changed from something like prettier (quotes, line breaks etc.)?

yoyo837:
Rebase please.

PR #580, opened 2019-07-02
Lock touch direction to avoid change on scroll

### LINKED ISSUE #308: Slider reacting to wrong gesture on touch devices

naikipl:
Fixes #308 

The `onStart` trigger is moved from `onTouchStart` to `onTouchMove` because we need to act on drag direction. The goal being: If the __first touch move__ is in the correct slider direction, trigger `onStart`; Otherwise, ignore further dragging.

Added `onTouchEnd` for cleanup.

As a result, the following user actions will no longer move a slider:
- touching the track
- dragging a horizontal slider vertically
- dragging a vertical slider horizontally

### DIFF
--- a/assets/index.less
+++ b/assets/index.less
@@ -27,7 +27,6 @@
   padding: 5px 0;
   width: 100%;
   border-radius: @border-radius-base;
-  touch-action: none;
   .borderBox();
 
   &-rail {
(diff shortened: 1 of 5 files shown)


yoyo837:
Rebase please.

PR #601, opened 2019-10-31
Update prop chart to include prefixCls

SebastianKurp:
Hey! 
Noticed that `prefixCls` is a prop that is exposed, but not documented. So I added it to the props chart in the readMe.
Here's an example code snippet of it in use, with a screenshot of the HTML to show that it is passing in the `prefixCls` to the className.
```
  <Slider
    prefixCls="predixClsTest"
  />
```
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ The following APIs are shared by Slider and Range.
 
 | Name         | Type    | Default | Description |
 | ------------ | ------- | ------- | ----------- |
+| prefixCls | string | `rc-tooltip` | Prefix class name |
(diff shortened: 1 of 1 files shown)


tlehtimaki:
This seems not to be the only one missing from the README.md. If you look at the src/Slider.tsx https://github.com/react-component/slider/blob/b5a337e71c6678f2d24e018509379a8e6f1d0518/src/Slider.tsx#L7-L47
The docs do not match with the interface very well or am I missing somehing? Maybe there should be a seperate issue and fix all at once.

yoyo837:
Rebase please.

PR #607, opened 2019-12-10
Fix keyboard control for null step with marks

weijiangan:
Currently, `util.calculateNextValue()` incorrectly returns mark object instead of value when steps are disabled (by passing `null` to `step` prop) and marks are used

The consequence of this is keyboard navigation will crash the slider due to unhandled errors that stem from unexpected value type.

# Steps to reproduce
1. Pass a marks object as prop to slider component
2. Pass ```step = {null}``` as prop to  slider component
3. Manipulate slider position using arrow keys


yoyo837:
Rebase please.

Issue #488, opened 2018-10-31
Range slider with center value and only one handle?


nardeas:
What would be the recommended way to achieve something like this:

![screen shot 2018-10-31 at 16 16 46](https://user-images.githubusercontent.com/11169240/47794175-9b602580-dd28-11e8-85e2-68a977a301fb.png)

![screen shot 2018-10-31 at 16 17 00](https://user-images.githubusercontent.com/11169240/47794185-a2873380-dd28-11e8-904f-802b6760040c.png)

I did implement that using some trickery like hiding the second handle, setting its value to 0 (center value) and swapping track and handle styles based on the second value but that produces an annoying side effect:

When the active handle crosses the "hidden" center handle the active one gets swapped to the previously hidden handle. Swapping styles works but the drag effect remains on the now hidden handle. You have to regrab the new handle to continue.

So is there a better way to do this with existing functionality that doesn't produce the sluggishness described?

nardeas:
Well I found out that matters how you hide the second handle. For example setting `display: none;` or `visibility: hidden;` produces the sluggishness. But if I set `background: transparent` and `border: transparent;` obviously it works.

All this seems a bit hacky to me, so the question remains: what would be the recommended way to achieve this?

photopaul25:
I am in need of the same behavior too. It would be greater if there was a variable that could be set for start position on a slider.

PR #609, opened 2019-12-16
trackStart prop for Slider component

iqqmuT:
By default the track starts from `min` value in Slider component. With new `trackStart` prop you can define where the track starts from.

Examples:
horizontal
vertical

See #488

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ The following APIs are shared by Slider and Range.
 | ------------ | ------- | ------- | ----------- |
 | defaultValue | number | `0` | Set initial value of slider. |
 | value | number | - | Set current value of slider. |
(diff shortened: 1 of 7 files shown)


zombieJ:
What's the best name of this prop? I have some concern.
Is that `centre` or `startPoint` or `basePoint` better?

@amoebageek  @iqqmuT  @afc163

iqqmuT:
It's surprisingly hard to come up with the right prop name. `startPoint` sounds good.

amoebageek:
> It's surprisingly hard to come up with the right prop name. `startPoint` sounds good.

Yes I agree, saying it startPoint is good

zombieJ:
@iqqmuT , could you help to update?

zombieJ:
`+ rc-slider@9.1.0`

PR #614, opened 2020-01-02
Do not trim-align values when props are changing

jeanlazarou:
When the value (ranges) are controlled (props) the trim-align process may change the new values in a wrong way, for instance [0, 60, 60, 240] produced [60, 60] breaking what the parent container actually passes.


yoyo837:
Rebase please.

PR #619, opened 2020-01-21
Fix floating-point arithmetic issues when props are decimals.

jcfrancisco:
Consider the following case:

```
      const value = 5.3;
      const props = {
        marks: {},
        step: 0.05,
        min: 0,
        max: 5.3
      };
```

I would expect the output of `getClosestPoint` to be `5.3`. However, it returns `5.25`.

This is because of floating point arithmetic issues with the function.

I've added a test case and taken a stab at figuring it out, making sure I continue to respect the `Math.floor` logic on `maxSteps`.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -19,19 +19,6 @@ export function isNotTouchEvent(e) {
     (e.type.toLowerCase() === 'touchend' && e.touches.length > 0);
 }
 
-export function getClosestPoint(val, { marks, step, min, max }) {
-  const points = Object.keys(marks).map(parseFloat);
-  if (step !== null) {
-    const maxSteps = Math.floor((max - min) / step);
-    const steps = Math.min((val - min) / step, maxSteps);
-    const closestStep =
-            Math.round(steps) * step + min;
-    points.push(closestStep);
-  }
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -19,28 +21,51 @@ export function isNotTouchEvent(e) {
...
 }
 
+export function getPrecision(step) {
+  const stepString = step.toString();
+  let precision = 0;
+  if (stepString.indexOf('.') >= 0) {
+    precision = stepString.length - stepString.indexOf('.') - 1;
+  }
+  return precision;
+}
```
jcfrancisco: This function is untouched, moved it up because I'm using it elsewhere.

src/utils.js
```diff
@@ -70,7 +95,7 @@ export function ensureValuePrecision(val, props) {
   const { step } = props;
   const closestPoint = isFinite(getClosestPoint(val, props)) ? getClosestPoint(val, props) : 0; // eslint-disable-line
   return step === null ? closestPoint :
-    parseFloat(closestPoint.toFixed(getPrecision(step)));
+    withPrecision(closestPoint, getPrecision(step));
```
jcfrancisco: I can back out this refactor if unneeded

PR #621, opened 2020-01-29
fix(#19879): focusing the wrong handle when using range

Kermit-Xuan:
[#19879](https://github.com/ant-design/ant-design/issues/19879)

### DIFF
--- a/assets/index.less
+++ b/assets/index.less
@@ -59,9 +59,12 @@
     background-color: #fff;
     touch-action: pan-x;
 
-    &:focus {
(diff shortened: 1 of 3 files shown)


Kermit-Xuan:
@afc163

Kermit-Xuan:
@afc163

zombieJ:
Could you help to add a test case for this patch?

Kermit-Xuan:
I don't see the test case about `allowCross` is `true` or drag.

Kermit-Xuan:
I have add test case for it

PR #625, opened 2020-02-13
Add parameter rootListener, which allows to restrict event listeners in bounds of specified html element.

tryvols:


zombieJ:
What's the api usage case?

tryvols:
@zombieJ Conflict of event listeners with scroll component from other library. They both used document, so the solution allows to override root element and make possible to resolve this conflict.

zombieJ:
Could you provide a reproduce for your case. I need more info to understand this. Thanks.

tryvols:
@zombieJ Reproduced https://codesandbox.io/s/polished-monad-ves9k

zombieJ:
> @zombieJ Reproduced https://codesandbox.io/s/polished-monad-ves9k

Thanks for reproduce. `stopPropagation` on DOM element **directly** is not controlled by React life cycle. You should always handle this in React way:

```jsx
<div onMouseMove={e => { e.stopPropagation(); }}>
  <AnyOtherComponent />
</div>
```

Please feel free to modify your reproduce with more reason that **MUST TO** handle event on DOM directly.

tryvols:
I made deeper research with trying to handle React event in the working project, but that's didn't working. In project we have situation, where we must use stopPropagation with just DOM, because there one lib, which broke scroll behavior using preventDefault in DOM. So, to prevent the handler, we must stopPropagation in DOM. Sorry, but I can't reproduce this situation, it's very complex. So we need possibility to set wrapper component of the slider as event listener, because event doesn't live on to the document element.

zombieJ:
Thanks for reply. You can fork repo to release a fork npm package since this api may not accept since we don't want to support this hack to support non-standard life cycle usage.

Issue #613, opened 2019-12-25
findDOMNode is deprecated in StrictMode


frontstall:
Hi guys, and thanks for this awesome component. React has started to warn recently about findDOMNode: Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Handle which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node

pethel:
https://github.com/react-component/slider/pull/627/files

mdrahiem:
Just curious to know why this pull is not merged yet? Coz I can see console error because of `findDOMNode is deprecated in StrictMode`.

pethel:
@mdrahiem 
Hi
It is not going to be merged either because it was a breaking change if the custom handle was used.

kruncher:
I've just added this package to a project and am seeing the same error with very simple usage:
```js
    <Slider
      min={ 0 }
      max={ 100 }
      value={ props.volume }
    />
```
Is the package currently unusable?

pethel:
For now, it is usable but with this annoying warning. In the future, I may not be, but hopefully, the team will fix it until then?

jecordjotse:
Any answer from the team about 'findDOMNode is deprecated in StrictMode' issue?

chrisdemetriad:
I am also waiting for this. What's going on?

cseas:
@zombieJ Any clue about whether this would be resolved? We're facing an issue in our project where rc-slider works fine in production mode but breaks the project's CSS in development mode. There are no errors thrown but I think it might have something to do with this.

zombieJ:
@cseas  what's the issue link? let me check.

carolineleger:
I still see the issue. Any news about it? Can the fix be merged?

cseas:
@zombieJ Sorry, false alarm, it was a problem with our project's bundler. But the "findDOMNode" console warning is still present so the original issue is still relevant.

SylannBin:
https://github.com/react-component/slider/issues/708

jacobbroughton:
Ran into this issue too, rip

ShoshaTheTarnegolet:
[...]

PR #627, opened 2020-02-19
refactor: removed deprecated findDOMNode

pethel:
When using React strict mode you get a warning about using findDOMNode. 

https://github.com/react-component/slider/issues/613


zombieJ:
This will break customize handle with out ref on the dom element directly:

https://github.com/ant-design/ant-design/blob/9a90c5b9101018f0ce4764c59b524adbe1f6e84d/components/slider/index.tsx#L110

yoyo837:
Ping @pethel

Issue #599, opened 2019-10-22
Range: Clicking on marks label will not update the slider


leventecsordas:
Clicking on the mark labels of a range slider will not update the slider. 

**Steps to reproduce:** 

1. got to the http://react-component.github.io/slider/examples/marks.html
2. find the `Range with marks` or `Range with marks and steps` sliders
3. click on any label

**Expected:** 

The slider should be updated, the nearest (or last changed) handler should jump on the above the label.

**Result:**

The slider is not updated, the handlers are not moved at all.

The problem as I can see is that the `value` passed from `onClickMarkLabel` is totally ignored and the bounds are not recalculated, so the `onChange` handler will be fired with previous `bounds`.

pdat:
I just updated from 8.6.1 to 8.7.1 and am now seeing this issue.

PR #637, opened 2020-03-13
Fix clicking on marks for range slider

maps82:
This solves https://github.com/react-component/slider/issues/599

A range can be changed by clicking the markers.


yoyo837:
Rebase please.

Issue #657, opened 2020-06-03
No export for RangeProps and HandleProps (9.3.0)


sbusch:
In our custom component based on `rc-slider`, we need access to the type definitions for `HandleProps` and `RangeProps`.

This worked in 9.2.4:

```typescript
import { Handle, HandleProps, Range, RangeProps } from 'rc-slider';
```

For 9.3.0, we now have to "circumvent" official exports and must directly import from internal files:

```typescript
import { Handle, Range } from 'rc-slider';
import { RangeProps } from 'rc-slider/lib/Range';
import { HandleProps } from 'rc-slider/lib/Handle';
```

This may apply also to `SliderProps` (which we don't use)

As this is an undocumented access, which may break in future, we kindly ask to re-export those types again. We would *love* to contribute a PR, but given the amount of open (unanswered) PRs, we will do this only if you give us the OK to proceed.

P.S issue #656 "Slider - cannot be used as a JSX component " also applies to us.

Kermit-Xuan:
It's unstable that depend on type definition file of Component because of it maybe be changed in the future.

I recommend this way to get type definition of Component:

``` ts
import * as React from 'react';
import { Range } from 'rc-slider';

type RangeProps = React.ComponentProps<typeof Range>;
```

cheyer:
Our team was also stumbling upon the missing type exports. @kerm1it is there any reason against adding them (like in @sbusch 's [PR](https://github.com/react-component/slider/pull/673))?

PR #673, opened 2020-08-20
add missing exports lost at release 9.3.0

sbusch:
see #657


kerm1it:
Trace in https://github.com/react-component/slider/pull/718

Issue #688, opened 2020-09-29
Range trackStyle and handleStyle do not accept an Array of CSSProperties


CesarSothen:
In the documentation says to use an Array of Objects when using trackStyle and handleStyle for Ranges.
When attempting to do it like so:
```
  <Range
        min={props.start}
        max={props.end}
        step={10}
        trackStyle={[{ height: "0.75rem", backgroundColor: "#4DADF5" }]}
        handleStyle={[
          {
            height: "1.5rem",
            width: "1.5rem",
          },
          {
            height: "1.5rem",
            width: "1.5rem",
          },
        ]}
        railStyle={{ height: "0.75rem" }}
        value={[props.values[0], props.values[1]]}
        onChange={(value: number[]) => props.handleChange(value[0], value[1])}
      />
```
It results in a compilation error because it does not support an array of objects.

To solve this had to change the Interface RangeProps from:
```
trackStyle?: React.CSSProperties;
handleStyle?: React.CSSProperties;
```
To:
```
trackStyle?: React.CSSProperties[];
handleStyle?: React.CSSProperties[];
```

PR #689, opened 2020-09-29
#688 - Fix to the Range track and handle styles

CesarSothen:
Fix to the issue #688 by changing the type of the RangeProps.


yoyo837:
Please help to fix antd/master lint error, after v9.5.2

yoyo837:
Why does this changes not merge into the minor version? @afc163

PR #718, opened 2020-12-01
chore: export slider、range and handle  props interface

kerm1it:
fix https://github.com/react-component/slider/issues/657

添加 SliderProps、RangeProps 和 HandleProps 导出


kerm1it:
@afc163 有空合了发个 patch

PR #724, opened 2020-12-14
Slider: Add dragging class when dragging

torkelo:
Wanted to customize handle style while dragging but could not find the dragging class being added for the Slider, only for the Range component.


yoyo837:
Rebase please.

PR #725, opened 2020-12-17
RangeProps allows number for `pushable`

RobertMurrayImprobable:
The top-level `README` describes the `pushable` prop of `Range` as `boolean` or `number` (or `undefined`), but the type annotation in `Range.tsx` is `pushable?: boolean;`. When developing with `rc-slider` in Typescript, using for example `pushable={5}` causes a type error, and the only way around this is `pushable={(5 as unknown) as undefined}` which then produces the expected behaviour. This type annotation appears to be incorrect and would be great to have it updated.

### DIFF
--- a/src/Range.tsx
+++ b/src/Range.tsx
@@ -39,7 +39,7 @@ export interface RangeProps extends GenericSliderProps {
   min?: number;
   max?: number;
   allowCross?: boolean;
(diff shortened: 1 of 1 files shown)


yoyo837:
Rebase please.

Issue #358, opened 2017-11-10
Missing key prop


schneckentempo:
I get the following error and would like to get this resolved:
![image](https://user-images.githubusercontent.com/1663926/32660004-5af0e81e-c621-11e7-838a-ca8eb8674781.png)

Looking into the code I think handles is missing the key prop

https://github.com/react-component/slider/blob/0a07adac3694f5881ecf9dffcd4831ba6fafeff9/src/Range.jsx#L307

yesmeck:
Could you try the latest version? Ping me if the issue still exists.

TkDodo:
The warning / error is still reproducable with version `8.6.9`.

Current code pointer seems to be this:

https://github.com/react-component/slider/blob/master/src/Range.jsx#L358

Could you re-open this issue?

PR #759, opened 2021-04-08
adds key to handle component

mateja176:
Reported in #358 

The issue was supposed to be fixed, nevertheless 7 people, including myself have witnessed the problem in the newest version (`9.7.2`).

With the following adjustment the issue was amended. Additionally the type signature of `Slider['handle']`:
```ts
handle?: (props: {
    className: string;
    prefixCls?: string;
    vertical?: boolean;
    offset: number;
    value: number;
    dragging?: boolean;
    disabled?: boolean;
    min?: number;
    max?: number;
    reverse?: boolean;
    index: number;
    tabIndex?: number;
    ariaLabel: string;
    ariaLabelledBy: string;
    ariaValueTextFormatter: string;
    style?: React.CSSProperties;
    ref?: React.Ref<any>;
  }) => React.ReactElement;
```

Already closely matches `React.FC`:
```ts
interface FunctionComponent<P = {}> {
        (props: PropsWithChildren<P>, context?: any): ReactElement<any, any> | null;
        propTypes?: WeakValidationMap<P>;
        contextTypes?: ValidationMap<any>;
        defaultProps?: Partial<P>;
        displayName?: string;
    }
```

With the accent on `(props: PropsWithChildren<P>, context?: any): ReactElement<any, any> | null;`. Hence there is no downside to instantiating it like a component.

Furthermore the presence of the `key` is supposed to save a number DOM updates, as `React` is able to match the instances by reference during reconciliation.

### DIFF
--- a/src/Range.tsx
+++ b/src/Range.tsx
@@ -495,7 +495,7 @@ class Range extends React.Component<RangeProps, RangeState> {
       min,
       max,
       reverse,
-      handle: handleGenerator,
+      handle: Handle,
       trackStyle,
       handleStyle,
       tabIndex,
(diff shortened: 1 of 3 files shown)


mateja176:
Closing due to inactivity

Issue #534, opened 2019-03-06
how to Disable hand icon on hover


skirankumar7:
how to Disable hand icon on hover

kkkasio:
you can pass this properties: 
ex: handleStyle={{ cursor: 'default' //none }}

saharme:
Added related PR:
https://github.com/react-component/slider/pull/766

PR #766, opened 2021-05-05
Added feature of setting the crosur to default rather than grab

saharme:
Solved issue:
https://github.com/react-component/slider/issues/534


yoyo837:
Rebase please.

Issue #654, opened 2020-05-29
RangeProps.pushable should have type boolean | number


jkkjonah:
Currently RangeProps prop `pushable` is typed as a boolean only. According to the documentation and source code it does support number values and should be updated accordingly

schoenwaldnils:
This is my workaround until this is fixed:

```js
const RangeWithFixedPushable = Range as React.ComponentClass<
  Omit<RangeProps, 'pushable'> & { pushable?: number | boolean }
>;
```

PR #782, opened 2021-06-18
fix(range): fixed pushable prop typing

### LINKED ISSUE #755: Pushable with Typescript
I tried to use pushable range. When I tried to set `pushable ` to number I get:

> Type "number" is not assignable to parameter of type "boolean | undefined"

Same thing when I tried use `RangeProps`. So I changed `pushable?: boolean;` in `Range.d.ts` to `pushable?: boolean | number;` and it works fine.
Am I doing right (I'm new to development, sorry)?

I'm using:
Typescript 4.1.2
React 17.0.1
rc-slider 9.7.1

lucasbordeau:
Added a **number** type on _pushable_ property of _Range_ component to reflect the real behavior of this property. 

See documentation : https://github.com/react-component/slider#range

Fixes #755 and #654

### DIFF
--- a/src/Range.tsx
+++ b/src/Range.tsx
@@ -39,7 +39,7 @@ export interface RangeProps extends GenericSliderProps {
   min?: number;
   max?: number;
   allowCross?: boolean;
-  pushable?: boolean;
+  pushable?: boolean | number;
   onChange?: (value: number[]) => void;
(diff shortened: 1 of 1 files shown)


fatso83:
How is this not being merged by now? Same as #725 and #713 which fixed #654

yoyo837:
conflicted

PR #791, opened 2021-07-23
feat: set exact step value on click

ronaldruzicka:
I needed a functionality to click on a step and set the exact value from it. Because even if I clicked on the step, it was actually clicking on the track, so it took the value were the cursor was and it was not always the same value as the step I clicked on, especially with small steps and wide range.

It would be nice to have it in your repo as well, let me know if it's ok like this.

Thanks.


yoyo837:
conflicted

PR #798, opened 2021-09-11
fix(utils): fix `ensureValuePrecision` to make sure value is always within range (min, max)

### LINKED ISSUE #797: The `handle` & `track` goes out while ensuring value precision
Given `min=-2.5` `max=2.5` `step=1`, `ensureValuePrecision` on edge values (`min` or `max`), it gives value out of scope like `-3` and `3`.

workaholicpanda:
fixes #797


yoyo837:
conflicted

PR #838, opened 2022-07-06
fix: page crash when dots is true and step is null

alanhaledc:
resolve ant-design/ant-design#36387


MadCcc:
Could you provide a test case?

PR #842, opened 2022-08-31
feat(dot-style): add marks dot classname

CookyG:
I need to do something special with the dot for marks, but I don't have a class name for marks, so I think it's worth adding that.


yoyo837:
Docs and test case please.

src/Steps/Dot.tsx
```diff
@@ -6,18 +6,23 @@ import SliderContext from '../context';
 export interface DotProps {
   prefixCls: string;
   value: number;
+  marksValue: number[];
```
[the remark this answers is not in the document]
yoyo837: We already have `dotStyle`, just add `dotClassName`.

src/Steps/Dot.tsx
```diff
@@ -33,9 +34,13 @@ export default function Dot(props: DotProps) {
...
   return (
     <span
-      className={classNames(dotClassName, {
-        [`${dotClassName}-active`]: active,
-      })}
+      className={classNames(
+        dotClassName,
+        {
+          [`${dotClassName}-active`]: active,
+        },
```
yoyo837, line 41:
```suggestion
          [`${dotClassName}-active`]: active,
```

src/Steps/index.tsx
```diff
@@ -23,6 +29,9 @@ export default function Steps(props: StepsProps) {
       dotSet.add(mark.value);
     });
 
+    //Fill marksValue
+    marksValueRef.current = Array.from(dotSet);
```
yoyo837, line 33: useEffect for `stepDots` change, or:

 const { stepDots,  marksValueRe }= React.useMemo(...)

CookyG: ok thanks！

src/Steps/index.tsx
```diff
@@ -32,20 +39,31 @@ export default function Steps(props: StepsProps) {
...
-      ))}
+      {stepDots.map((dotValue) => {
+        // Check whether it is a marks dot
+        const isMarksDot = marksValue.indexOf(dotValue) >= 0;
+
+        return (
+          <Dot
+            prefixCls={prefixCls}
+            className={classNames({
+              [marksDotClassName]: isMarksDot,
```
yoyo837, line 58: Need test case for `className`

yoyo837, line 58:
```suggestion
              [`${prefixCls}-marks-dot`]: isMarksDot,
```

src/Steps/index.tsx
```diff
@@ -32,20 +36,31 @@ export default function Steps(props: StepsProps) {
...
-      ))}
+      {stepDots.map((dotValue) => {
+        // Check whether it is a marks dot
+        const isMarksDot = marksValue.indexOf(dotValue) >= 0;
+
+        return (
+          <Dot
+            prefixCls={prefixCls}
+            className={classNames({
+              [`${prefixCls}-marks-dot`]: isMarksDot,
```
yoyo837, line 55: Add css for `${prefixCls}-marks-dot`  in here https://github.com/react-component/slider/blob/master/assets/index.less

yoyo837, line 55:
```diff
[`${prefixCls}-marks-dot`]: isMarksDot,
+markedDot,
```

@MadCcc An additional?

MadCcc: Yeah, current design do not need a className to style Slider. It's better to append customized className to this element.

CookyG: OK, I'll write it down when I have time.

PR #866, opened 2023-04-04
perf: useMemo for marks and steps

madzgo:
This is a performance improvement that stops Marks and Steps components from re-rendering while dragging handles/track.


yoyo837:
Seems to contain a bit of unnecessary changes.

PR #904, opened 2023-05-23
fix(allowCross): When two handles coincide (i.e pushNum = 0 || pushab…

nag-murali:
…le = false, allowCross === false ) and you can move the handle across the other handle which should not be the case


yoyo837:
Please add some test case for it.

afc163:
图片

Please add some test case for it.

PR #945, opened 2023-09-21
feat: Move xxStyle & xxClassName into styles & classNames

zombieJ:


luna-hibachi-xyz:
How do we specify colors for different segments now if `TrackStyles `is gone? I saw style.track cannot accept an array

PR #949, opened 2023-11-09
Allow easier styling of mark and dot

razzeee:
Similar to the other components we have like `tracks`, `rail` and `handle`, this allows easier styling via classes or inline styles.

I've also separated the active styles of both, as being able to style, how a selected range should look is very helpful.

### DIFF
--- a/src/Marks/Mark.tsx
+++ b/src/Marks/Mark.tsx
@@ -1,7 +1,7 @@
+import cls from 'classnames';
 import * as React from 'react';
-import classNames from 'classnames';
-import { getDirectionStyle } from '../util';
 import SliderContext from '../context';
+import { getDirectionStyle } from '../util';
 
 export interface MarkProps {
   prefixCls: string;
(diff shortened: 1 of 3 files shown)


razzeee:
This might also allow to drop `dotStyle` and `activeDotStyle` from the top level, like seems to have been done for the other styles

src/Marks/Mark.tsx
```diff
@@ -13,22 +13,30 @@ export interface MarkProps {
...
+          [`${textCls}-active`]: active,
+        },
+        active && classNames.markActive,
+        classNames.mark,
+      )}
       style={{
         ...positionStyle,
         ...style,
+        ...(active ? styles.markActive : {}),
+        ...styles.mark,
```
lvnam96, line 39: This should be:
```javascript
...styles.mark,
...(active ? styles.markActive : {}),
```
`markActive` styles should override the default `mark` styles, not the other way around.

src/interface.ts
```diff
@@ -2,11 +2,23 @@ import type React from 'react';
 
 export type Direction = 'rtl' | 'ltr' | 'ttb' | 'btt';
 
-export type OnStartMove = (e: React.MouseEvent | React.TouchEvent, valueIndex: number, startValues?: number[]) => void;
+export type OnStartMove = (
+  e: React.MouseEvent | React.TouchEvent,
+  valueIndex: number,
+  startValues?: number[],
+) => void;
```
lvnam96, line 9: Unnecessary changes.

lvnam96 (changes requested):

PR #952, opened 2023-11-29
fix: onAfterChange should not trigger when click on track

MadCcc:
close https://github.com/ant-design/ant-design/issues/46158


src/Slider.tsx
```diff
@@ -329,7 +329,6 @@ const Slider = React.forwardRef((props: SliderProps, ref: React.Ref<SliderRef>)
 
       onBeforeChange?.(getTriggerValue(cloneNextValues));
       triggerChange(cloneNextValues);
-      onAfterChange?.(getTriggerValue(cloneNextValues));
```
yoyo837, line 332: 为啥 onBeforeChange 可以有， onAfterChange不能？

MadCcc: onAfterChange 必须对应到 mouseup 事件上

Issue #736, opened 2021-01-06
How to use a custom svg as the handle


cseas:
I'm trying to use a custom svg icon as the handle for my slider but the svg doesn't appear at all.

Here's what I tried:

```tsx
import PauseIcon from "./pause.svg";
...
      <Slider
        handleStyle={{
          borderColor: "white",
          height: 32,
          width: 32,
          marginTop: -13,
          boxShadow: "0 2px 6px 0 rgba(0, 0, 0, 0.28)",
          background: PauseIcon,
        }}
      />
```

cseas:
Figured it out from discussion in #449

Here's how:

```tsx
import { ReactComponent as PauseIcon } from "./pause.svg";
...
      <Slider
        handleStyle={{
          borderColor: "white",
          height: 32,
          width: 32,
          marginTop: -13,
          boxShadow: "0 2px 6px 0 rgba(0, 0, 0, 0.28)",
        }}
        handle={(handleProps: any) => {
          return (
            <Handle {...handleProps}>
                <PauseIcon width="10px" />
            </Handle>
          );
        }}
      />
```

AlfredMadere:
This does not work in 2023. Handle is not exported from "rc-slider" and the handle prop does not exist on the Slider component. Can this still be achieved?

UmbraCi:
> This does not work in 2023. Handle is not exported from "rc-slider" and the handle prop does not exist on the Slider component. Can this still be achieved?

use V9 version，V10 has be refactored. some mistake should update

yoyo837:
See https://github.com/react-component/slider/pull/963#discussion_r1423618617

PR #963, opened 2023-12-11
Add custom element as handle

MajesticKoala:
Added customHandle to Slider props allowing the user to pass in a custom element which will be displayed as the handle.

Aims to solve needing to downgrade to V9 as mentioned here: https://github.com/react-component/slider/issues/736#issuecomment-1563991591

### DIFF
--- a/src/Handles/Handle.tsx
+++ b/src/Handles/Handle.tsx
@@ -24,6 +24,7 @@ export interface HandleProps {
   onBlur?: (e: React.FocusEvent<HTMLDivElement>) => void;
   render?: (origin: React.ReactElement<HandleProps>, props: RenderProps) => React.ReactElement;
   onChangeComplete?: () => void;
+  customHandle?: React.JSX.Element;
 }
 
 const Handle = React.forwardRef<HTMLDivElement, HandleProps>((props, ref) => {
@@ -37,6 +38,7 @@ const Handle = React.forwardRef<HTMLDivElement, HandleProps>((props, ref) => {
     dragging,
     onOffsetChange,
     onChangeComplete,
+    customHandle,
     ...restProps
   } = props;
   const {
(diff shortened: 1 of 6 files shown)


yoyo837:
Please add some test case for it.

MajesticKoala:
https://github.com/react-component/slider/assets/31652524/5ce7f6a8-ff32-477e-8a0b-ef4d992b8ac4

src/Handles/Handle.tsx
```diff
@@ -162,7 +165,15 @@ const Handle = React.forwardRef<HTMLDivElement, HandleProps>((props, ref) => {
       aria-valuetext={getIndex(ariaValueTextFormatterForHandle, valueIndex)?.(value)}
       aria-orientation={direction === 'ltr' || direction === 'rtl' ? 'horizontal' : 'vertical'}
       {...restProps}
-    />
+    >
+      {customHandle && (
```
yoyo837, line 169: Handle accepts the `render` prop function, which seems to be able to do the same thing. Are they redundant to each other?

MajesticKoala: It is possible, seems easiest to clone the element and add custom components as children. Although the styles of the handleNode will still apply to children so have to either manually set or set visiblity.

```
const customRender = (handle: React.ReactElement) => {
  return React.cloneElement(handle, {
    style: {
      ...handle.props.style,
      visibility: 'hidden',
    }
  }, 
    <CustomHandle /> // Add style 'visiblity': 'visible' to child component
  );
};
```
And simply passing in 
`handleRender={customRender}`

Happy to close the PR

yoyo837: Ok, let's close it.

janKir: Just ran into this and am a little surprised that this seems to be the supposed way to customize the slider handle.

The [docs on React.cloneElement](https://react.dev/reference/react/cloneElement) state: "Pitfall: Using cloneElement is uncommon and can lead to fragile code." So I would favor to have a more straight-forward solution to custom handle elements. Otherwise, it would be nice to have at least some documentation and examples. But seems like the README is not up to date to the v10 changes.

PR #1089, opened 2026-08-26
fix: finish drag when mouseup propagation stops

### LINKED ISSUE #1024: OnMouseUp never called if <Slider> is a child of a parent component with an "OnMouseUp"
When a <Slider> is added as a child of a parent element, and that parent element has an 'onMouseUp' event handler that stops propagation, the slider continues to call its "onChange" event after the user stops interacting with the Slider itself.

This problem is reproduced here:
https://codesandbox.io/p/sandbox/rc-slider-mouseup-cvx3j8

nrps9909:
## Summary

- observe drag-ending `mouseup` in the capture phase;
- remove the listener with the same capture option in both normal completion and unmount cleanup;
- add a regression for a slider inside a parent that stops mouseup propagation.

## Background

Fixes #1024.

The drag listener was attached to `document` in the bubble phase. If any ancestor handled `mouseup` and called `stopPropagation()`, the event never reached that listener. The slider therefore remained in its dragging transaction: `onChangeComplete` was not called, the document `mousemove` listener stayed attached, and later pointer movement continued changing the value after release.

Listening for the document-level end event in the capture phase lets the slider close its own drag transaction before an ancestor intentionally stops bubbling. The listener registration and both removal paths use the same capture flag.

Exact-base reproduction: the new regression left `onChangeComplete` at zero calls. With this patch it receives the final dragged value, and a subsequent document mousemove produces no `onChange` call.

## Verification

- `npm test -- --runInBand`: 5 suites, 120 tests, 5 snapshots passed
- `npm run tsc`: passed
- `npm run lint`: passed
- `npm run compile`: ESM, CJS, declarations, and CSS passed
- `git diff --check`: passed

AI assistance disclosure: Codex was used to trace event propagation and cleanup paths, audit duplicate PRs, draft the focused regression, and run the verification above. The final diff and results were reviewed locally before submission.

## Summary by CodeRabbit

* **Bug 修复**
  * 修复父级阻止鼠标抬起事件冒泡时，Slider 拖拽无法正常结束的问题。
  * 拖拽结束后将正确停止响应后续鼠标移动，并触发完成回调。
  * 优化鼠标与触摸操作的事件处理，避免不必要地阻止浏览器默认行为。


src/hooks/useDrag.ts
```diff
@@ -213,7 +213,7 @@ function useDrag(
     const onMouseUp: EventListener = (event) => {
       event.preventDefault();
```
yoyo837: Now that this runs in the capture phase, `preventDefault()` executes before consumer `onMouseUp` handlers, changing `defaultPrevented` from `false` to `true`. Please keep `preventDefault()` only for `touchend` to preserve the existing mouseup behavior.

nrps9909: Fixed in signed commit 88b4752: mouseup now preserves the existing defaultPrevented=false behavior, while touchend retains preventDefault. I extended the propagation regression to assert the parent mouseup handler observes defaultPrevented=false. The focused Slider file passes 55 tests; the full suite passes 122 tests and 5 snapshots, together with TypeScript, lint, Prettier, ESM/CJS/declaration and CSS compilation, and diff validation.

nrps9909:
Follow-up signed commit a536a73 covers the touchend-only preventDefault branch that Codecov identified. The regression starts a real touch drag, dispatches touchend, verifies the native default is prevented exactly once, and confirms onChangeComplete receives the unchanged range. Focused: 51 Range tests pass. Full coverage run: 123 tests and 5 snapshots pass; useDrag.ts now reports the new line covered. TypeScript and lint pass, and the touched files are Prettier-clean.

